PR #1, opened 2021-05-20
add a shell script framework to support clusteradm init and join

mikeshng:
- framework/skeleton only. 
- copied and modified from https://github.com/open-cluster-management/cluster-keeper
- init and join sub-commands are added but doesn't have real implementation

Sample outputs currently:
```
$ ./clusteradm 
usage: clusteradm [OPTIONS] SUBCOMMAND

    SUBCOMMAND is one of the following.

    init
        Starting the open-cluster-management control plane
    join
        Initializes a klusterlet agent on the managed cluster and joins it to the hub cluster.

    The following OPTIONS are available:

    -h    Display usage for SUBCOMMAND
    -v    Verbosity level for information printed to stderr. Default: 0
```
init
```
$ ./clusteradm -h init
usage: clusteradm init [CONTEXT]

    Starting the open-cluster-management control plane

    CONTEXT is the name of a kubeconfig context
```
```
./clusteradm init
Kubernetes control plane is running at https://127.0.0.1:33467
KubeDNS is running at https://127.0.0.1:33467/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
```
join
```
./clusteradm -h join
usage: clusteradm join [CONTEXT]

    Initializes a klusterlet agent on the managed cluster and joins it to the hub cluster.

    CONTEXT is the name of a kubeconfig context
```

Signed-off-by: Mike Ng <ming@redhat.com>


KevinFCormier:
You might want to look into whether there are some frameworks out there for generating Bash CLIs. Everything in cluster-keeper was from scratch just because I was in a rush to get something to help us out with using ClusterPools and I didn't take the time to really research it. Didn't really anticipate it would get as popular as it seems to be. There are some limitations and repetitive boilerplate in the way things are structured, and the order of options and arguments is not very flexible. (Options always have to come right after the command or subcommand, depending which they apply to.)

mikeshng:
Thanks for the heads up Kevin.  I think we should be ok for now despite the limitation you mentioned. The shell/bash implementation is just a temporary solution to gather some feedback. We will most likely move to a golang implementation in the near future.

PR #5, opened 2021-05-26
Add version and fix README/CONTRIBUTE

itdove:


README.md
```diff
@@ -1,7 +1,7 @@
 [comment]: # ( Copyright Contributors to the Open Cluster Management project )
 # clusteradm CLI & CLI Plugin
 
-A CLI and kubernetes CLI plugin that allows you to interact with OCM/ACM to provision and managed your Hybrid Cloud presence from the command-line.
+A CLI and kubernetes CLI plugin that allows you to interact with open-cluster-management to provision and managed your Hybrid Cloud presence from the command-line.
```
qiujian16, line 4: s/provision and managed/manage/

pkg/cmd/version/exec.go
```diff
@@ -0,0 +1,79 @@
...
+	fmt.Printf("client\t\tversion\t:%s\n", clusteradm.GetVersion())
+	client, err := helpers.GetControllerRuntimeClientFromFlags(o.ConfigFlags)
+	if err != nil {
+		return err
+	}
+	return o.runWithClient(client)
+}
+
+func (o *Options) runWithClient(client crclient.Client) (err error) {
+	cms := &corev1.ConfigMapList{}
```
qiujian16, line 37: let's empty  this function. I think it tries to get image-manfest in ACM which will not work.

sonar-project.properties
```diff
@@ -1,5 +1,5 @@
-sonar.projectKey=open-cluster-management_cm-cli
-sonar.projectName=cm-cli
+sonar.projectKey=open-cluster-management_clusteradm
```
qiujian16, line 1: let's remove sonar file until we decide whether we also need to integrate sonar here.

README.md
itdove: done

pkg/cmd/version/exec.go
itdove: it works, if image-manifest not present the list is empty and so jump here: https://github.com/itdove/clusteradm/blob/bd9c173b4231918e5a3928d85e590c584cb8ce6f/pkg/cmd/version/exec.go#L68

on kind:

(base) dvernier@dvernier-mac ~ % clusteradm version
client		version	:0.0.1
server release	version	:v1.19.7

As per the arch call this morning, I understood that we have to do function that works for both products.

sonar-project.properties
itdove: Done

pkg/cmd/version/exec.go
itdove: There is also a e2e test on it which run on kind: https://github.com/itdove/clusteradm/blob/6ca5416204fd1b207082bc103695a03232adbe32/build/run-functional-tests.sh#L20

qiujian16: i think we should remove wording like `acm-custom-registry` from the code. 
If the intention is to fetch the server version. We need  a  better approach instead  of reading it from image manifest...So it is better to make this an empty function at first.

itdove: I will just remove the ACM part of it

qiujian16:
thanks please squash the commit, then we can merge this.

itdove:
Can't squash and merge in GitHub directly ?

qiujian16:
I haven't enable squash in openshift-merge-bot yet....And I would want openshift-merge-bot to merge PRs....

itdove:
Close as to complex to squash, will open a new one.

PR #6, opened 2021-06-02
Init cmd

itdove:


pkg/cmd/init/exec.go
```diff
@@ -0,0 +1,91 @@
...
+
+	err = helpers.ApplyDeployment(kubeClient, reader, scenarioDirectory, o.values, "init/operator.yaml")
+	if err != nil {
+		return err
+	}
+	err = helpers.ApplyCustomResouces(dynamicClient, reader, scenarioDirectory, o.values, "init/clustermanagers.cr.yaml")
+	if err != nil {
+		return err
+	}
+
```
qiujian16, line 76: followup: add watch here to monitor the status of clustermanager.

pkg/cmd/init/options.go
```diff
@@ -0,0 +1,29 @@
...
+	factory     cmdutil.Factory
+	values      Values
+}
+
+type Values struct {
+	Hub Hub `json:"hub"`
+}
+
+type Hub struct {
+	TokenID     string `json:"tokenID"`
```
qiujian16, line 20: all these would need comment if they are public variables

pkg/cmd/init/exec.go
```diff
@@ -0,0 +1,91 @@
...
+	apiServerInternal, err := helpers.GetAPIServer(kubeClient)
```
qiujian16, line 77: i would think we move this to join command. So in join command, you construct a client to reach to hub and get this url.

pkg/cmd/init/scenario/init/cluster_role_binding_bootstrap.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: cluster-bootstrap
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:open-cluster-management:bootstrap
+subjects:
+- kind: Group
+  apiGroup: rbac.authorization.k8s.io
+  name: system:bootstrappers:managedcluster
```
qiujian16, line 13: eof

pkg/cmd/init/scenario/init/bootstrap-token-secret.yaml
```diff
@@ -0,0 +1,19 @@
...
+stringData:
+  # Token ID and secret. Required.
+  token-id: {{ .Hub.TokenID }}
+  token-secret: {{ .Hub.TokenSecret }}
+
+  # Allowed usages.
+  usage-bootstrap-authentication: "true"
+
+  # Extra groups to authenticate the token as. Must start with "system:bootstrappers:"
+  auth-extra-groups: system:bootstrappers:managedcluster
```
qiujian16, line 19: eof

pkg/cmd/init/options.go
```diff
@@ -0,0 +1,29 @@
...
+	TokenSecret string `json:"tokenSecret"`
+}
+
+func newOptions(f cmdutil.Factory, streams genericclioptions.IOStreams) *Options {
+	return &Options{
+		ConfigFlags: genericclioptions.NewConfigFlags(true),
+		factory:     f,
+	}
+}
```
qiujian16, line 29: followup: we should be able to set image repository/tag as flags in init command.

pkg/cmd/init/cmd.go
```diff
@@ -0,0 +1,54 @@
...
+
+var valuesTemplatePath = filepath.Join(scenarioDirectory, "values-template.yaml")
+
+// NewCmd ...
+func NewCmd(f cmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command {
+	o := newOptions(f, streams)
+
+	cmd := &cobra.Command{
+		Use:          "init",
+		Short:        "init hub",
```
qiujian16, line 33: why init hub? can it be just init?

pkg/cmd/init/scenario/init/service_account.yaml
```diff
@@ -0,0 +1,6 @@
+# Copyright Contributors to the Open Cluster Management project
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cluster-manager
+  namespace: open-cluster-management
```
qiujian16, line 6: eof

pkg/cmd/join/scenario/join/bootstrap_hub_kubeconfig.yaml
```diff
@@ -0,0 +1,9 @@
+# Copyright Contributors to the Open Cluster Management project
+apiVersion: v1
+kind: Secret
+metadata:
+  name: bootstrap-hub-kubeconfig
+  namespace: open-cluster-management-agent
+type: Opaque
+data:
+  kubeconfig: {{ .Hub.KubeConfig | b64enc }}
```
qiujian16, line 9: eof

pkg/cmd/join/exec.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return err
+	}
+
+	clientUnSecure, err := kubernetes.NewForConfig(restConfigUnSecure)
+	if err != nil {
+		return err
+	}
+
+	ca, err := helpers.GetCACert(clientUnSecure)
```
qiujian16, line 104: you can  get apiserver and ca from here. 

I think the logic would be
1. get apiserver-endpoint in cluster-info at first
2. if it is not found, use apiserver from the --hub-apiserver.

In this way, you do not need another flag --hub-internal-apiserverr.

pkg/cmd/join/exec.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	bootstrapConfig := bootstrapConfigUnSecure
+	bootstrapConfig.Clusters[0].Cluster.InsecureSkipTLSVerify = false
+	bootstrapConfig.Clusters[0].Cluster.CertificateAuthorityData = ca
+	bootstrapConfig.Clusters[0].Cluster.Server = o.values.Hub.InternalServerURL
+	bootstrapConfigBytes, err := yaml.Marshal(bootstrapConfig)
+	if err != nil {
+		return err
+	}
```
qiujian16, line 116: let's move the part to generate bootstrap kubeconfig into a separate func

pkg/cmd/join/options.go
```diff
@@ -0,0 +1,35 @@
...
+package join
+
+import (
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+type Options struct {
+	ConfigFlags       *genericclioptions.ConfigFlags
+	token             string
```
qiujian16, line 11: should we have a description for every flags?

pkg/helpers/apply.go
```diff
@@ -0,0 +1,177 @@
...
+		if err != nil {
+			return err
+		}
+		gvk := gvks[0]
+		resource, err := getResource(u.GetKind())
+		if err != nil {
+			return err
+		}
+		dr := client.Resource(gvk.GroupVersion().WithResource(resource))
+		if u.GetNamespace() != "" {
```
qiujian16, line 102: you do not need this if. client will not add ns in the path if there is none

pkg/helpers/apply.go
```diff
@@ -0,0 +1,177 @@
...
+			_, err = dr.Namespace(u.GetNamespace()).
+				Create(context.TODO(), u, metav1.CreateOptions{})
```
qiujian16, line 104: if this is an apply func, you would need to also handle update.

qiujian16 (commented):
thanks, i think in general it looks good. I have some minor comment.
I think we would need a check in init/join command, to ensure it can only be run once?

We need some follow up maybe in another pr.

pkg/helpers/apply.go
itdove: Done

pkg/helpers/apply.go
itdove: Done

pkg/cmd/join/exec.go
itdove: Done... hope it is what you expected

pkg/cmd/init/cmd.go
```diff
@@ -0,0 +1,54 @@
...
+		Example:      fmt.Sprintf(example, helpers.GetExampleHeader()),
```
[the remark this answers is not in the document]
itdove: https://github.com/open-cluster-management-io/clusteradm/issues/7

pkg/cmd/accept/exec.go
```diff
@@ -0,0 +1,161 @@
...
+			if _, err := signingRequest.UpdateApproval(context.TODO(), csr.Name, csr, metav1.UpdateOptions{}); err != nil {
+				return err
+			}
+
+			fmt.Printf("CSR %s approved\n", csr.Name)
+		} else {
+			fmt.Printf("no CSR to approve for cluster %s\n", clusterName)
+		}
+
+		mc := &unstructured.Unstructured{}
```
qiujian16, line 133: I would prefer to use type here instead of unstructured. If you are using client from controller-runtime. I think you can add the  scheme in it and use the cluster type. I would want this as a followup.

pkg/helpers/apply.go
```diff
@@ -94,17 +98,27 @@ func ApplyCustomResouces(client dynamic.Interface,
...
+		mapper := restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(discoveryClient))
+		mapping, err := mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
 		if err != nil {
 			return err
 		}
-		dr := client.Resource(gvk.GroupVersion().WithResource(resource))
-		if u.GetNamespace() != "" {
-			_, err = dr.Namespace(u.GetNamespace()).
-				Create(context.TODO(), u, metav1.CreateOptions{})
+		// resource, err := getResource(u.GetKind())
```
qiujian16, line 106: remove these code which are commented.

qiujian16 (commented):
/lgtm

Thanks, I leave some minor comments which will not block merging the code.

PR #14, opened 2021-06-08
Implement --dry-run / --output-file

itdove:

### DIFF
--- a/pkg/cmd/init/options.go
+++ b/pkg/cmd/init/options.go
@@ -3,17 +3,15 @@ package init
 
 import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
-	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	genericclioptionsclusteradm "open-cluster-management.io/clusteradm/pkg/genericclioptions"
 )
 
 //Options: The structure holding all the command-line options
 type Options struct {
-	//ConfigFlags: The generic options from the kubernetes cli-runtime.
-	ConfigFlags *genericclioptions.ConfigFlags
-	factory     cmdutil.Factory
-	values      Values
-	//if set the resources will be sent to stdout instead of being applied
-	dryRun bool
+	//ClusteradmFlags: The generic optiosn from the clusteradm cli-runtime.
+	ClusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags
+	// factory         cmdutil.Factory
+	values Values
 	//The file to output the resources will be sent to the file.
 	outputFile string
 }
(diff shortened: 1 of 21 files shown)


pkg/cmd/init/options.go
```diff
@@ -12,6 +12,10 @@ type Options struct {
 	ConfigFlags *genericclioptions.ConfigFlags
 	factory     cmdutil.Factory
 	values      Values
+	//if set the resources will be sent to stdout instead of being applied
+	dryRun bool
```
qiujian16, line 16: I think you can make it as a generic option instead of define it in every command.

pkg/cmd/join/exec.go
```diff
@@ -100,32 +101,37 @@ func (o *Options) run() error {
...
-	err = apply.ApplyDeployment(kubeClient, reader, o.values, "", "join/operator.yaml")
+	out, err = apply.ApplyDeployment(kubeClient, reader, o.values, o.dryRun, "", "join/operator.yaml")
 	if err != nil {
 		return err
 	}
+	output = append(output, out...)
 
-	b := retry.DefaultBackoff
-	b.Duration = 100 * time.Millisecond
+	if !o.dryRun {
```
qiujian16, line 116: a better way is you have

```
if o.dryRun {
  output things
  return
}

apply manifests
```

pkg/helpers/apply/apply.go
```diff
@@ -45,82 +47,109 @@ func ApplyDeployment(
...
 func ApplyDirectly(clients *resourceapply.ClientHolder,
 	reader asset.ScenarioReader,
 	values interface{},
+	dryRun bool,
 	headerFile string,
-	files ...string) error {
+	files ...string) ([]string, error) {
 	recorder := events.NewInMemoryRecorder(helpers.GetExampleHeader())
+	output := make([]string, 0)
+	for _, name := range files {
```
qiujian16, line 93: I would move this into a separate func such as `renderManifests`, So if dry-run is enabled, I will call it.

pkg/cmd/join/exec.go
itdove: no I can not return because I still want to capture the output for the next ApplyCustomResouces.
I just want to skip the WaitCRDToBeReady as the CRD will not be created because it is dryRun.

pkg/cmd/init/options.go
itdove: I will see what I can do to create genericOption for the clusteradm

pkg/helpers/apply/apply.go
itdove: ok

qiujian16 (commented):
/lgtm

Issue #16, opened 2021-06-11
bootstrap token is not supported on all *KS


itdove:
The boostrap token which is generated while initializing the hub is not supported by every *KS and thus the spoke fails to join.

In such case a service-account must be created and the sa token use as bootstrap token.

PR #17, opened 2021-06-13
Use service-account to join if not --use-bootstrap-token on init

itdove:
https://github.com/open-cluster-management-io/clusteradm/issues/16
/hold


pkg/helpers/client.go
```diff
@@ -30,7 +32,7 @@ func GetAPIServer(kubeClient kubernetes.Interface) (string, error) {
 	return cluster.Server, nil
 }
 
-func GetCACert(kubeClient kubernetes.Interface) ([]byte, error) {
+func GetCACert(kubeClient kubernetes.Interface, useBootstrapToken bool) ([]byte, error) {
```
qiujian16, line 35: do you need useBootstrapToken?

pkg/cmd/init/exec.go
```diff
@@ -92,18 +120,31 @@ func (o *Options) run() error {
...
-		o.values.Hub.TokenSecret,
+		token,
 		restConfig.Host,
 	)
 
 	return apply.WriteOutput(o.outputFile, output)
 }
+
+func waitForBootstrapSecret(kubeClient kubernetes.Interface, b wait.Backoff) (secret *corev1.Secret, err error) {
+	err = retry.OnError(b, func(err error) bool {
```
qiujian16, line 139: i might use watch with fiedselector here. But it is a blocker.

pkg/helpers/client.go
itdove: You are correct it is not needed anymore

pkg/cmd/init/exec.go
itdove: I don't know how to use watch as it is not a controller. Do you mean it is a blocker or not?

qiujian16: not blocker, sorry. Updated the comment

qiujian16 (commented):
/lgtm

PR #24, opened 2021-06-17
Avoid re-init and add get token

itdove:
https://github.com/open-cluster-management-io/clusteradm/issues/21
https://github.com/open-cluster-management-io/clusteradm/issues/22
/hold


build/run-functional-tests.sh
```diff
@@ -84,5 +84,16 @@ else
...
+then
+   echo "get token command result: "$CMDINITRESULT
+   ERROR_REPORT=$ERROR_REPORT+"clusteradm get token failed\n"
+else
+   echo "get token command result: "$CMDINITRESULT
+   echo $CMDINITRESULT
+fi
+
 kind delete cluster --name $CLUSTER_NAME-hub
 kind delete cluster --name $CLUSTER_NAME-spoke
```
qiujian16, line 99: eof

itdove: Done

pkg/cmd/get/token/exec.go
```diff
@@ -0,0 +1,184 @@
...
+		secret, err = helpers.GetBootstrapSecret(kubeClient)
+		return err
+	})
+	return
+}
+
+func getToken(kubeClient kubernetes.Interface) (string, error) {
+	var bootstrapSecret *corev1.Secret
+	saSecret, err := helpers.GetBootstrapSecret(kubeClient)
+	if err != nil {
```
qiujian16, line 162: move 164-178 to a separate func. I think it will be reused.

pkg/cmd/get/token/exec.go
```diff
@@ -0,0 +1,184 @@
...
+				if strings.HasPrefix(s.Name, config.BootstrapSecretPrefix) {
+					bootstrapSecret = &s
+				}
+			}
+			if bootstrapSecret != nil {
+				return fmt.Sprintf("%s.%s", string(bootstrapSecret.Data["token-id"]), string(bootstrapSecret.Data["token-secret"])), nil
+			}
+		}
+		return "", err
+	} else {
```
qiujian16, line 181: else is not needed here

pkg/cmd/get/token/exec.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return err
+	}
+
+	clientHolder := resourceapply.NewClientHolder().
+		WithAPIExtensionsClient(apiExtensionsClient).
+		WithKubernetes(kubeClient).
+		WithDynamicClient(dynamicClient)
+
+	token, err := getToken(kubeClient)
```
qiujian16, line 86: add a comment that you want to retrieve token from bootstrap or sa.

pkg/cmd/get/token/exec.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
```
qiujian16, line 87: I would write a 
```
func applyToken(userbootstrap, dryrun) (string, error)
```

pkg/cmd/get/token/exec.go
```diff
@@ -0,0 +1,184 @@
...
+		if err != nil {
+			return err
+		}
+		output = append(output, out...)
+		if !o.ClusteradmFlags.DryRun {
+			//Make sure that the sa token is ready
+			if !o.useBootstrapToken {
+				b := retry.DefaultBackoff
+				b.Duration = 100 * time.Millisecond
+				_, err = waitForBootstrapSecret(kubeClient, b)
```
qiujian16, line 117: I do not think it will work, since you apply sa with dryrun, so no secret will be generated.

pkg/cmd/get/token/exec.go
```diff
@@ -0,0 +1,184 @@
...
+	}, func() error {
+		secret, err = helpers.GetBootstrapSecret(kubeClient)
+		return err
+	})
+	return
+}
+
+func getToken(kubeClient kubernetes.Interface) (string, error) {
+	var bootstrapSecret *corev1.Secret
+	saSecret, err := helpers.GetBootstrapSecret(kubeClient)
```
qiujian16, line 161: this func name GetBootstrapSecret is confusing, i would change to GetBootstrapSecretFromSA

pkg/cmd/get/token/exec.go
```diff
@@ -0,0 +1,184 @@
...
+				}
+			}
+			token, err = getToken(kubeClient)
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		//Update the cluster role
+		files := []string{
```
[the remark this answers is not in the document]
itdove: This is the issue that happened today.
I requested the token once, it created the sa and role/binding. Then we changed the role because we were missing something the "create". Requesting again the token, didn't updated the role and so wasn't working. I added the 'else' so at least the role is updated every time if needed.

pkg/config/env.go
```diff
@@ -5,4 +5,8 @@ package config
 const (
 	OpenClusterManagementNamespace = "open-cluster-management"
 	BootstrapSAName                = "cluster-bootstrap"
+	ClusterManagerName             = "cluster-manager"
+	LabelApp                       = "app"
+	LabelAppClusterApp             = "cluster-manager"
```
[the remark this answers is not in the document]
itdove: You ALWAYS set the Label values as the operator name?

pkg/cmd/init/scenario/init/bootstrap_cluster_role.yaml
```diff
@@ -28,3 +28,9 @@ rules:
   - create
   - list
   - update
+- apiGroups: 
+  - "cluster.open-cluster-management.io"
+  resources: 
+  - "managedclustersets/join"
+  verbs: 
+  - "create"
```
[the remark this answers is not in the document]
itdove: Done

pkg/helpers/client.go
```diff
@@ -120,3 +120,18 @@ func GetBootstrapSecret(
 	}
 	return secret, nil
 }
+
+func IsClusterManagerInstalled(apiExtensionsClient apiextensionsclient.Interface) (bool, error) {
```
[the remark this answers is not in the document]
itdove: Done

pkg/config/env.go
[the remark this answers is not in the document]
qiujian16: I do not think we or user will change that.

pkg/cmd/get/token/exec.go
itdove: Done

pkg/cmd/get/token/exec.go
itdove: Done

pkg/cmd/get/token/exec.go
itdove: done

pkg/cmd/get/token/exec.go
itdove: That's why there is a if to skip it if dryRun

pkg/cmd/get/token/exec.go
itdove: Done

pkg/config/env.go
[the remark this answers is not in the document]
itdove: Done

pkg/cmd/get/token/exec.go
```diff
@@ -0,0 +1,164 @@
...
+	clientHolder := resourceapply.NewClientHolder().
+		WithAPIExtensionsClient(apiExtensionsClient).
+		WithKubernetes(kubeClient).
+		WithDynamicClient(dynamicClient)
+
+	//Retrieve token from service-account and if not found
+	//from the bootstrap token
+	token, err := helpers.GetToken(kubeClient)
+	if err != nil {
+		if !errors.IsNotFound(err) {
```
qiujian16, line 87: let's to like this

```go
switch {
case errors.IsNotFound(err):
    out, err := o.applyToken(clientHolder, reader)
    output = append(output, out...)	
    if err != nil {
        return err
    }
case err != nil:
    return err
}

if o.ClusteradmFlags.DryRun {
    return func() err {
        fmt.Printf("token=%s\n", token)
        fmt.Printf("please log on spoke and run:\n%s join --hub-token %s --hub-apiserver %s --cluster-name &lt;cluster_name&gt;\n",		helpers.GetExampleHeader(),		token,		restConfig.Host,	)
        return apply.WriteOutput(o.outputFile, output)
    }()
}

if o.useBootstrapToken {
   // apply clusterrole
   return  func() err {
        fmt.Printf("token=%s\n", token)
        fmt.Printf("please log on spoke and run:\n%s join --hub-token %s --hub-apiserver %s --cluster-name &lt;cluster_name&gt;\n",		helpers.GetExampleHeader(),		token,		restConfig.Host,	)
        return apply.WriteOutput(o.outputFile, output)
    }()
}

// do thing with sa token
```

itdove: I think the GetToken method is not a good idea as it searches first the SA and then the bootstrap-token which make difficult to find a solution for all scenarios (use bootstrap token when SA already installed and vice versa, when no token is created yet and you stick with the same kind of token it wokrs well). So instead of calling GetToken, I call GetBootstrapToken or GetBootstrapTokenFromSA depending on the useBootstrapToken

qiujian16 (commented):
/lgtm

qiujian16:
/hold

would you squash it? it has 12 commits in one pr...

itdove:
I will do tomorrow, thx

Issue #25, opened 2021-06-18
Add a wait with timeout on the `clusteradm accept`


itdove:
This should wait (with timeout) that each CSRs and Managedcluster from the list of provided clusters are present before approving/updating them.

PR #26, opened 2021-06-19
Add wait for accept

itdove:
https://github.com/open-cluster-management-io/clusteradm/issues/25

### DIFF
--- a/pkg/cmd/accept/exec.go
+++ b/pkg/cmd/accept/exec.go
@@ -5,9 +5,13 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"open-cluster-management.io/clusteradm/pkg/helpers"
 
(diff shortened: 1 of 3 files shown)


pkg/cmd/accept/exec.go
```diff
@@ -61,95 +64,133 @@ func (o *Options) run() error {
...
-					done = true
-					break
-				}
+		}
+		//Does not contain the correct group
+		if len(group) == 0 {
+			continue
+		}
+		//Check if already approved or denied
+		for _, c := range item.Status.Conditions {
```
qiujian16, line 129: would you move this to a seperate IsCSRApproved func? like this
https://github.com/open-cluster-management-io/registration/blob/e83faae5f8b9646e0244f180839e38e186dcffc4/pkg/clientcert/certificate.go#L161

pkg/cmd/accept/exec.go
```diff
@@ -61,95 +64,133 @@ func (o *Options) run() error {
...
-				kubeClient, err := o.ClusteradmFlags.KubectlFactory.KubernetesClientSet()
-				if err != nil {
-					return err
-				}
-				signingRequest := kubeClient.CertificatesV1().CertificateSigningRequests()
-				if _, err := signingRequest.UpdateApproval(context.TODO(), csr.Name, csr, metav1.UpdateOptions{}); err != nil {
-					return err
-				}
+	if csr != nil {
+		if !o.ClusteradmFlags.DryRun {
```
qiujian16, line 144: this seems weird....I think the whole func should not be called with dryrun?

pkg/cmd/accept/exec.go
```diff
@@ -61,95 +64,133 @@ func (o *Options) run() error {
...
-			if !o.ClusteradmFlags.DryRun {
-				mc.Spec.HubAcceptsClient = true
-				_, err = clusterClient.ClusterV1().ManagedClusters().Update(context.TODO(), mc, metav1.UpdateOptions{})
-				if err != nil {
-					return err
-				}
+		return false, err
+	}
+	if !mc.Spec.HubAcceptsClient {
+		if !o.ClusteradmFlags.DryRun {
```
qiujian16, line 184: ditto

pkg/cmd/accept/exec.go
itdove: No in dry-run,  we will not approve but at a least we will check if csr is present.

pkg/cmd/accept/exec.go
itdove: same for managed cluste r... we check if the managedcluster is present but in dry-run we don't update it.

pkg/cmd/accept/exec.go
itdove: yes can do

itdove: but will be isApprovedOrDenied

pkg/cmd/accept/exec.go
qiujian16: in that case i think you can do
```go
if o.ClusteradmFlags.DryRun {
  return true, nil
}
if !mc.Spec.HubAcceptsClient {
  ....
}
```

pkg/cmd/accept/exec.go
```diff
@@ -61,95 +64,133 @@ func (o *Options) run() error {
...
+	}
+	var csr *certificatesv1.CertificateSigningRequest
+	for _, item := range csrs.Items {
+		//Does not have the correct name prefix
+		if !strings.HasPrefix(item.Spec.Username, userNameSignaturePrefix) {
+			continue
+		}
+		//Check groups
+		var group string
+		for _, g := range item.Spec.Groups {
```
qiujian16, line 118: i would use
```go
groups := sets.NewString(item.Spec.Groups...)
if !groups.Has(groupName) {
  continue
}
```

pkg/cmd/accept/exec.go
itdove: but the link... doesn't check if already denied... if already denied should we give a chence to approve it again?

pkg/cmd/accept/exec.go
qiujian16: then it might be

```
if csr == nil {
    fmt.Printf("no CSR to approve for cluster %s\n", clusterName)
    return false, nil
}
if o.ClusteradmFlags.DryRun {
    return true, nil
}

// approve csr
```

pkg/cmd/accept/exec.go
qiujian16: if it is already denied, it cannot be approve anymore

itdove: then my code is good, because the example you gave still allow an approve after get denied,

pkg/cmd/accept/exec.go
itdove: ok more simple ;-)

pkg/cmd/accept/exec.go
itdove: yes

pkg/cmd/accept/exec.go
qiujian16: my point is move those part into a separate func

pkg/cmd/accept/exec.go
itdove: if I do that, the message `httpAcceptsClient already set for cluster` will be not display in dry-run

qiujian16: try this 
```go
if mc.Spec.HubAcceptsClient {
  fmt.Printf("httpAcceptsClient already set for cluster %s\n", clusterName)
  return true, nil
}
if o.ClusteradmFlags.DryRun {
  return true, nil
}

//approve

```

qiujian16 (approved):
just a nit comment, and we are good to go thanks!

/approve

PR #28, opened 2021-06-24
Revoke token

itdove:
https://github.com/open-cluster-management-io/clusteradm/issues/23 add `clusteradm delete token`

/hold


pkg/cmd/delete/token/exec.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	//Delete Roles
+	err = kubeClient.RbacV1().ClusterRoles().Delete(context.TODO(), config.BootstrapClusterRoleName, metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	//Detele bootstrap token secret
+	secret, err := helpers.GetBootstrapSecret(kubeClient)
```
qiujian16, line 76: why get  it?

qiujian16:
/approve

just a nit comment.

pkg/cmd/delete/token/exec.go
itdove: Because the name is like: `bootstrap-token-{{ .Hub.TokenID }}` and so we don't know the exact name, we have to search it to delete it.

PR #32, opened 2021-06-26
Add a customFuncMap parameter

itdove:
I added that capabability because I needed for the cm_cli to overload the template functions.


pkg/helpers/apply/apply.go
```diff
@@ -45,6 +45,7 @@ var (
 func ApplyDeployments(
 	kubeClient kubernetes.Interface,
 	reader asset.ScenarioReader,
+	customFuncMap template.FuncMap,
```
qiujian16, line 48: I am not sure the purpose of this. Is it to render the yaml template in a customized way?

itdove:
It is to be able to add new functions in the Gand text/template rendering. For the time being there is the standard golang text functions plus spring, plus few others I implemented in the templatefuntion.go file. I want to implement few extra for the cm-cli but as they are specific to the cm cli I don't want to implement them in the clusteradm. This change will allow me to inject the extra functions I need for the cm cli without having them in the clusteradm

pkg/helpers/apply/applierBuilder.go
```diff
@@ -0,0 +1,55 @@
...
+	templateFuncMap     template.FuncMap
+}
+
+//ApplierBuilder a builder to build the applier
+type ApplierBuilder struct {
+	Applier
+}
+
+type iApplierBuilder interface {
+	Build() Applier
```
qiujian16, line 26: Add some comments for each method here.

qiujian16:
it looks good. Would you squash the commit.

PR #43, opened 2021-07-05
Add some logs

itdove:
Other logs will come along.

### DIFF
--- a/pkg/cmd/accept/exec.go
+++ b/pkg/cmd/accept/exec.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
 	"open-cluster-management.io/clusteradm/pkg/helpers"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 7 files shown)


pkg/cmd/accept/exec.go
```diff
@@ -70,10 +72,14 @@ func (o *Options) run() error {
 func (o *Options) runWithClient(kubeClient *kubernetes.Clientset, clusterClient *clusterclientset.Clientset) (err error) {
 	for _, clusterName := range o.values.clusters {
 		if o.wait == 0 {
-			_, err = o.accept(kubeClient, clusterClient, clusterName)
+			var csrApproved bool
+			csrApproved, err = o.accept(kubeClient, clusterClient, clusterName, false)
+			if err == nil && !csrApproved {
+				err = fmt.Errorf("no CSR to approve for cluster %s", clusterName)
```
qiujian16, line 78: we did not check case with err != nil here...And the output message is here with line 150. You may want to add more details?

itdove: yes we check err != nil at like 85.
The line 150 is displayed only when waitMode...  so it display continuesly until either the wait time is expired or the accept succeed.

qiujian16: I see, thanks

qiujian16 (commented):
/approve
/lgtm

PR #61, opened 2021-08-24
make the description message more clear

xauthulei:
it is a bit confuse to the newer when creating log file. Thanks


xauthulei:
Thanks @qiujian16

PR #66, opened 2021-09-20
With owner

itdove:


pkg/helpers/apply/apply_suite_test.go
```diff
@@ -0,0 +1,249 @@
...
+			applier := applierBuilder.
+				WithClient(kubeClient, apiExtensionsClient, dynamicClient).
+				WithTemplateFuncMap(FuncMap()).WithOwner(managedClusterOwner, false, false, scheme.Scheme).
+				Build()
+
+			_, err := applier.ApplyCustomResources(reader, nil, false, "", "ownerref/managedcluster.yaml")
+			Expect(err).To(BeNil())
+		})
+		By("Checking Ownerref", func() {
+			mc, err := clusterClient.ClusterV1().ManagedClusters().Get(context.TODO(), "my-managedcluster", metav1.GetOptions{})
```
qiujian16, line 142: This seems to add a onwer of one mc to another?

itdove: yes it is just and example, I used 2 mc, could be one mc and something else. it doesn't matter which CRD I use... it is just to test that one resource becomes the owner of another.

PR #72, opened 2021-10-27
application lifecycle refactoring related changes

mikeshng:
clusteradm related changes for app lifecycle refactoring https://github.com/open-cluster-management-io/multicloud-operators-subscription/pull/35 

/hold

Signed-off-by: Mike Ng <ming@redhat.com>


pkg/cmd/install/addons/scenario/addons/appmgr/clusterrole.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: multicloud-operators
+  name: multicluster-operators
```
qiujian16, line 5: change this to a more specific name. I think it is only used for subscription

mikeshng: You are right. Done as suggested, changed all `multicluster-operators` references to `multicluster-operators-subscription`. Thanks.

qiujian16:
/approve
/lgtm
/hold

hold in case you need other to review. And I would REALLY love to see placement replaces placementrule as a followup :)

mikeshng:
> I would REALLY love to see placement replaces placementrule as a followup :)

Based on discussion with @xiangjingli and @jnpacker we should have a concrete target date for adopting placement API by this Friday.

> hold in case you need other to review.

There are no one else that needs to review but I did need to update a yaml file (crd_report.yaml). Please take a quick look again, thanks.

PR #80, opened 2021-12-08
Feat: Allows override OCM images upon initialization

yue9944882:

### DIFF
--- a/pkg/cmd/init/cmd.go
+++ b/pkg/cmd/init/cmd.go
@@ -21,8 +21,10 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	o := newOptions(clusteradmFlags, streams)
 
 	cmd := &cobra.Command{
-		Use:          "init",
-		Short:        "init the hub",
+		Use:   "init",
+		Short: "Initialize a Kubernetes cluster into an OCM hub cluster.",
+		Long: "Initialize the Kubernetes cluster in the context into an OCM hub cluster by applying a few" +
+			"fundamental resources including registration-operator, etc.",
 		Example:      fmt.Sprintf(example, helpers.GetExampleHeader()),
 		SilenceUsage: true,
 		PreRun: func(c *cobra.Command, args []string) {
@@ -46,5 +48,9 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	cmd.Flags().StringVar(&o.outputFile, "output-file", "", "The generated resources will be copied in the specified file")
 	cmd.Flags().BoolVar(&o.useBootstrapToken, "use-bootstrap-token", false, "If set then the boostrap token will used instead of a service account token")
 	cmd.Flags().BoolVar(&o.force, "force", false, "If set then the hub will be reinitialized")
+	cmd.Flags().StringVar(&o.registry, "image-registry", "quay.io/open-cluster-management",
+		"The name of the image registry serving OCM images.")
+	cmd.Flags().StringVar(&o.version, "version", "latest",
+		"The installing version of OCM components.")
 	return cmd
 }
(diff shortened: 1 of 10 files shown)


qiujian16:
/assign @ilan-pinto
/approve

This looks ok to me in general

ilan-pinto:
general comments:
1. you assume that all the repo and versions are the same. in the init you have 4 images. it might be ok, i would recommend pointing that out in the Flag
2.missing functional tests

pkg/cmd/init/exec.go
```diff
@@ -23,8 +23,10 @@ func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
 	klog.V(1).InfoS("init options:", "dry-run", o.ClusteradmFlags.DryRun, "force", o.force, "output-file", o.outputFile)
 	o.values = Values{
 		Hub: Hub{
-			TokenID:     helpers.RandStringRunes_az09(6),
-			TokenSecret: helpers.RandStringRunes_az09(16),
+			TokenID:           helpers.RandStringRunes_az09(6),
+			TokenSecret:       helpers.RandStringRunes_az09(16),
+			ReleaseVersion:    o.version,
```
ilan-pinto, line 28: move ReleaseVersion ImageRegistryName to another struct

ilan-pinto (changes requested):

pkg/cmd/init/cmd.go
```diff
@@ -46,5 +48,9 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	cmd.Flags().StringVar(&o.outputFile, "output-file", "", "The generated resources will be copied in the specified file")
 	cmd.Flags().BoolVar(&o.useBootstrapToken, "use-bootstrap-token", false, "If set then the boostrap token will used instead of a service account token")
 	cmd.Flags().BoolVar(&o.force, "force", false, "If set then the hub will be reinitialized")
+	cmd.Flags().StringVar(&o.registry, "image-registry", "quay.io/open-cluster-management",
+		"The name of the image registry serving OCM images.")
+	cmd.Flags().StringVar(&o.version, "version", "latest",
```
itdove, line 53: I would call this one tag instead as version.

pkg/cmd/init/exec.go
```diff
@@ -49,6 +51,12 @@ func (o *Options) validate() error {
 	if installed {
 		return fmt.Errorf("hub already initialized")
 	}
+	if len(o.registry) == 0 {
```
itdove, line 54: As there is a default, this is not needed to check o.registry nor o.version

pkg/cmd/init/options.go
```diff
@@ -17,6 +17,10 @@ type Options struct {
 	useBootstrapToken bool
 	//if true the hub will be reinstalled
 	force bool
+	//Installing release version of OCM
+	version string
```
itdove, line 21: would call it tag @qiujian16 What do you think

pkg/cmd/init/exec.go
yue9944882: factored out a `Image` struct

pkg/cmd/init/cmd.go
yue9944882: renamed to tag

pkg/cmd/init/exec.go
yue9944882: but i guess it's fine to validate them in case of unexpected misconfiguration?

pkg/cmd/init/options.go
yue9944882: renamed to tag

yue9944882:
> you assume that all the repo and versions are the same. in the init you have 4 images. it might be ok, i would recommend pointing that out in the Flag

i clarified it in the subcommand's "usage"

pkg/cmd/init/options.go
qiujian16: I would call it version since the tag could be different for images.

yue9944882:
> 2.missing functional tests

mmm i think we'd beer testify the functionalities via e2e blackbox test against real kind sandbox clusters in the github action, wdyt? i can add an e2e in a follow-up b/c a new e2e framework needs to be initiated in the repo.

qiujian16:
one missing part is to specify the placement image which tag is different from registration and work. I think we may need to maintain a internal map from version to image tags...

I think we can merge this one at first. I will create an issue to add placement image in cluster manager api and also factoring version flags.

yue9944882:
> 2.missing functional tests

covered the new flag in the integration test scripts

yue9944882:
> one missing part is to specify the placement image which tag is different from registration and work. I think we may need to maintain a internal map from version to image tags...

the placement image will be defaulted to latest if missing, and placement release version (v0.2.0) doesnt align to the overall OCM release version (v0.5.0). i suppose it's fine to keep it latest in the current phase

PR #81, opened 2021-12-08
Feat: Adding an install scripts for one line clusteradm installation

yue9944882:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,196 @@
+# Copyright Contributors to the Open Cluster Management project
+#!/usr/bin/env bash
+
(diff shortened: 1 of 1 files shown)


yue9944882:
now the clusteradm installation should work by:

> /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/yue9944882/clusteradm/feat/install-sh/install.sh)"

qiujian16:
/approve
/lgtm

btw, I think we can use `go get open-cluster-management.io/clusteradm` also ^^

PR #86, opened 2021-12-13
Add fossa.yml

itdove:


mikeshng:
@qiujian16 `FOSSA_API_KEY` needs to be added as the secret to this repo (or just add it as a secret for the whole org). I don't have permission to modify the settings for this repo or org. Thanks.

PR #90, opened 2021-12-21
Feat: Clusteradm proxy health command

yue9944882:

/cc @qiujian16 

the new command basically works by running a local port-forward server for the proxy-server in the hub cluster, and calling `/healthz` endpoint for each of the managed cluster.

a sample of the new command is:

the happy path:
```
CLUSTER NAME    INSTALLED    AVAILABLE    PROBED HEALTH    LATENCY
managed1        True         True         True             67.595144ms
zuoxiu-kind     True         True         True             85.418368ms
```

if anything goes wrong in a managed cluster:
```shell
E1220 19:08:19.836955   37375 exec.go:159] Failed requesting /healthz endpoint for cluster zuoxiu-kind
CLUSTER NAME    INSTALLED    AVAILABLE    PROBED HEALTH    LATENCY
managed1        True         True         True             82.003112ms
zuoxiu-kind     True         False        False            <timeout>
```

furthremore, we can use the local proxy client certificates by setting flags `--proxy-ca-cert/--proxy-cert/--proxy-key`.


pkg/cmd/proxy/probe/exec.go
```diff
@@ -0,0 +1,324 @@
...
+		Get(context.TODO(), clusterAddon.Spec.AddOnConfiguration.CRName, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "failed getting managedproxyconfiguration for cluster-proxy")
+	}
+
+	clusterClient, err := clusterv1.NewForConfig(hubRestConfig)
+	if err != nil {
+		return errors.Wrapf(err, "failed newing cluster client")
+	}
+	managedClusterList, err := clusterClient.ManagedClusters().List(
```
qiujian16, line 111: maybe list related managedclusteraddon here? Since a user may not start proxy in some clusters.

pkg/cmd/proxy/probe/exec.go
```diff
@@ -0,0 +1,324 @@
...
+	certSecret, err := nativeClient.CoreV1().Secrets(proxyConfig.Spec.ProxyServer.Namespace).
+		Get(context.TODO(), inClusterSecretClient, metav1.GetOptions{})
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed getting cert & key secret")
+	}
+	certData := certSecret.Data["tls.crt"]
+	keyData := certSecret.Data["tls.key"]
+
+	// building tls config from pem-encoded data
+	tlsCfg, err := buildTLSConfig(caData, certData, keyData, "127.0.0.1", nil)
```
qiujian16, line 203: why it is 127.0.0.1?

pkg/cmd/proxy/probe/exec.go
```diff
@@ -0,0 +1,324 @@
...
+		Get(context.TODO(), common.AddonName, metav1.GetOptions{})
+	installed := "True"
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return errors.Wrapf(err, "failed getting managed cluster addon for cluster %v", clusterName)
+		}
+		installed = "False"
+	}
+	addonAvailable := "False"
+	for _, cond := range addon.Status.Conditions {
```
qiujian16, line 228: nit, use meta.IsStatusConditionTrue

pkg/cmd/proxy/probe/exec.go
```diff
@@ -0,0 +1,324 @@
...
+		if cond.Type == addonv1alpha1.ManagedClusterAddOnConditionAvailable {
+			addonAvailable = string(cond.Status)
+		}
+	}
+
+	copiedCfg := rest.CopyConfig(hubRestConfig)
+	copiedCfg.Dial = dialFunc
+	copiedCfg.TLSClientConfig = rest.TLSClientConfig{}
+	copiedCfg.TLSClientConfig.Insecure = true
```
qiujian16, line 237: maybe add a comment or TODO here

pkg/cmd/proxy/probe/exec.go
```diff
@@ -0,0 +1,324 @@
...
+	if !isFailed {
+		end := time.Now()
+		data, _ := ioutil.ReadAll(resp.Body)
+		if "ok" == string(data) {
+			health = "True"
+			latency = end.Sub(start).String()
+		}
+	}
+
+	w.print(clusterName, installed, addonAvailable, health, latency)
```
qiujian16, line 273: TODO use a common table convertor in the future.

pkg/cmd/proxy/probe/exec.go
```diff
@@ -0,0 +1,324 @@
...
+	latency := "<none>"
+	start := time.Now()
+	isFailed := false
+	resp, err := rt.RoundTrip(req)
+	if err != nil {
+		isFailed = true
+		klog.Errorf("Failed requesting /healthz endpoint for cluster %v", clusterName)
+		if strings.Contains(err.Error(), "dial timeout") {
+			latency = "<timeout>"
+		}
```
qiujian16, line 262: I think we can print here and return directly, so it is not needed to check isFailed

pkg/cmd/proxy/probe/exec.go
yue9944882: ah oversight

pkg/cmd/proxy/probe/exec.go
yue9944882: ideally yes, but for now the automatic cluster-proxy addon installation cannot be opt-out due to framework's limitation. so every managed cluster will always install an addon constantly, what do you think if we adding the filtering after that is supported in the framework?

pkg/cmd/proxy/probe/exec.go
yue9944882: updated

pkg/cmd/proxy/probe/exec.go
yue9944882: adding a TODO comment and elaboration about why insecure TLS is needed here

pkg/cmd/proxy/probe/exec.go
yue9944882: removed isFailed check

pkg/cmd/proxy/probe/exec.go
yue9944882: added

qiujian16 (commented):
/approve
/lgtm

PR #92, opened 2021-12-22
Bugfix: Fixes go install due to replace directive

yue9944882:


yue9944882:
@qiujian16 i moved the dirname from `cmd` to `clusteradm` on purpose, otherwise the installed binary in the gopath will be named `cmd` instead of `clusteradm`

qiujian16 (commented):
/approve
/lgtm

Also need to change the related doc on website

PR #93, opened 2021-12-24
Feat: Clusteradm init supports saving commands to local file for future use

yue9944882:


qiujian16 (commented):
should we just regenerate this command?

yue9944882:
https://github.com/open-cluster-management-io/cluster-proxy/pull/34

@qiujian16 actually i'm trying to re-use the generated command from the e2e job, does that make sense?

qiujian16:
I think it makes sense, but it seems we try to maintain how to combine join command in two places.

yue9944882:
@qiujian16 now the generated commands are converged

qiujian16 (commented):
/approve
/lgtm

PR #94, opened 2021-12-27
Feat: Clusteradm init --wait for progressive hub init

yue9944882:


yue9944882:
/hold

until https://github.com/open-cluster-management-io/clusteradm/pull/93, will rebase after the precedence merged

pkg/cmd/init/exec.go
```diff
@@ -140,10 +149,29 @@ func (o *Options) run() error {
...
+			return err
+		}
+	}
+
 	restConfig, err := o.ClusteradmFlags.KubectlFactory.ToRESTConfig()
 	if err != nil {
 		return nil
 	}
+	if len(o.outputJoinCommandFile) > 0 {
+		sh, err := os.OpenFile(o.outputJoinCommandFile, os.O_CREATE|os.O_WRONLY, 0755)
```
qiujian16, line 165: Should have a flag to only output this command so we can pipe it to a file?

yue9944882: are you thinking about utilizing the output from init e.g.?

> echo "$JOIN_CLUSTER_NAME" | cluster init --cmd-only --pipe-stdin

pkg/cmd/join/exec.go
```diff
@@ -193,7 +193,7 @@ func waitUntilRegistrationOperatorConditionIsTrue(f util.Factory, timeout int64)
 	operatorSpinner.Start()
 	defer operatorSpinner.Stop()
 
-	return waitUntilConditionIsTrue(
+	return helpers.WatchUntil(
```
qiujian16, line 196: Followup: We should use "degraded" condition of klusterlet here instead of reading pod status.

qiujian16 (commented):
/approve
/lgtm

PR #100, opened 2022-01-04
Bugfix: Fixes --wait flag condition and test flakiness

yue9944882:
this will also fix the test flakiness in https://github.com/open-cluster-management-io/clusteradm/pull/99


qiujian16 (commented):
/approve
/lgtm

Issue #87, opened 2021-12-15
more command to manage addons


qiujian16:
We can to `clusteradm addon enable` today to enable a certain addon. It is valuable to also

1. list all addons on hub
2. list all enabled addon on a managedcluster
3. disable an addon on a cluster

qiujian16:
I think this are almost done. Thanks @ycyaoxdu !

/close

PR #102, opened 2022-01-05
Feat: clusteradm addon disable command for disable specified addon

ycyaoxdu:
Signed-off-by: ycyaoxdu yaoyuchen0626@163.com

Add a subcommand `diable` to command `addon` for enhance the new feature 3 which mentioned in  https://github.com/open-cluster-management-io/clusteradm/issues/87

### DIFF
--- a/pkg/cmd/addon/disable/cmd.go
+++ b/pkg/cmd/addon/disable/cmd.go
@@ -45,9 +45,8 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 		},
 	}
 
-	cmd.Flags().StringVar(&o.names, "name", "", "Names of the add-on to deploy (comma separated)")
-	cmd.Flags().StringVarP(&o.namespace, "namespace", "n", "open-cluster-management-agent-addon", "Specified namespace to addon addon")
-	cmd.Flags().StringVar(&o.clusters, "cluster", "", "Names of the managed cluster to deploy the add-on to (comma separated)")
+	cmd.Flags().StringSliceVar(&o.names, "name", []string{}, "Names of the add-on to deploy (comma separated)")
+	cmd.Flags().StringSliceVar(&o.clusters, "cluster", []string{}, "Names of the managed cluster to deploy the add-on to (comma separated)")
 
 	return cmd
 }
--- a/pkg/cmd/addon/disable/cmd.go
+++ b/pkg/cmd/addon/disable/cmd.go
@@ -1,0 +1,53 @@
+// Copyright Contributors to the Open Cluster Management project
+package disable
+
+import (
+	"fmt"
+
+	genericclioptionsclusteradm "open-cluster-management.io/clusteradm/pkg/genericclioptions"
+	clusteradmhelpers "open-cluster-management.io/clusteradm/pkg/helpers"
+
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+var example = `
+# Disanable addon on a cluster in speccified a namespace
+%[1]s addon disable --name application-manager --ns namespace --cluster cluster1,cluster2
+`
+
+// NewCmd...
+func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, streams genericclioptions.IOStreams) *cobra.Command {
+	o := newOptions(clusteradmFlags, streams)
+
+	cmd := &cobra.Command{
+		Use:          "disable",
+		Short:        "disable specified addon on specified managed cluster",
+		Example:      fmt.Sprintf(example, clusteradmhelpers.GetExampleHeader()),
+		SilenceUsage: true,
+		PreRunE: func(c *cobra.Command, args []string) error {
+			clusteradmhelpers.DryRunMessage(clusteradmFlags.DryRun)
+
+			return nil
+		},
+		RunE: func(c *cobra.Command, args []string) error {
+			if err := o.complete(c, args); err != nil {
+				return err
+			}
+			if err := o.validate(); err != nil {
+				return err
+			}
+			if err := o.run(); err != nil {
+				return err
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&o.names, "name", "", "Names of the add-on to deploy (comma separated)")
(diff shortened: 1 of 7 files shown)


ycyaoxdu:
something goes wrong. https://github.com/open-cluster-management-io/clusteradm/runs/4836202667?check_suite_focus=true#step:4:65 
It seems like a bug ,need to be solved.
 i will check it out first.

ycyaoxdu:
solved by this change https://github.com/open-cluster-management-io/clusteradm/pull/107/files#diff-57fb64e41aecc769c85303b36a57d69f497e6ba8390f3352a4f1485a029ca1d1R17 

> something goes wrong. https://github.com/open-cluster-management-io/clusteradm/runs/4836202667?check_suite_focus=true#step:4:65 It seems like a bug ,need to be solved. i will check it out first.

pkg/cmd/addon/disable/exec.go
```diff
@@ -0,0 +1,117 @@
...
+		return fmt.Errorf("clusters is misisng")
+	}
+	return nil
+}
+
+func (o *Options) run() (err error) {
+	alreadyProvidedAddons := make(map[string]bool)
+	addons := make([]string, 0)
+	names := strings.Split(o.names, ",")
+	for _, n := range names {
```
qiujian16, line 43: you can use sets in apimachinery here.

pkg/cmd/addon/disable/cmd.go
```diff
@@ -0,0 +1,53 @@
...
+			if err := o.run(); err != nil {
+				return err
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&o.names, "name", "", "Names of the add-on to deploy (comma separated)")
+	cmd.Flags().StringVarP(&o.namespace, "namespace", "n", "open-cluster-management-agent-addon", "Specified namespace to addon addon")
```
qiujian16, line 49: I do not think the ns and cluster here is different. You only need cluster. For `enable`, it should be the same.

pkg/cmd/addon/disable/exec.go
```diff
@@ -0,0 +1,117 @@
...
+		if _, ok := alreadyProvidedAddons[n]; !ok {
+			alreadyProvidedAddons[n] = true
+			addons = append(addons, strings.TrimSpace(n))
```
qiujian16, line 46: why append here? I think the code is easier if using sets.

pkg/cmd/addon/disable/cmd.go
ycyaoxdu: ok ,will also remove that in `enable`

pkg/cmd/addon/disable/exec.go
qiujian16: you can just use addons:= sets.NewString(names...)

pkg/cmd/addon/disable/exec.go
```diff
@@ -0,0 +1,114 @@
...
+	names := strings.Split(o.names, ",")
+	for _, n := range names {
+		if !addons.Has(strings.TrimSpace(n)) {
+			addons.Insert(strings.TrimSpace(n))
+		}
+	}
+
+	clusters := sets.NewString()
+	cs := strings.Split(o.clusters, ",")
+	for _, c := range cs {
```
qiujian16, line 51: clusters := sets.NewString(cs...)

pkg/cmd/addon/disable/cmd.go
```diff
@@ -0,0 +1,52 @@
...
+				return err
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&o.names, "name", "", "Names of the add-on to deploy (comma separated)")
+	cmd.Flags().StringVar(&o.clusters, "cluster", "", "Names of the managed cluster to deploy the add-on to (comma separated)")
+
```
qiujian16, line 50: I think we should have a`--all-clusters` command that disables addons on all clusters as a followup.

We may also do the same for enable command.

pkg/cmd/addon/disable/exec.go
qiujian16: btw, could you ensure if we did the same for enable command?

pkg/cmd/addon/disable/exec.go
ycyaoxdu: Will this change leads to repeated element, space-prefixed element or space-suffixed element to be inserted into the set if we just use `cs`, which is result of `o.clusters` split by `,`, to initialize the set. 
For example, if the input string is `--clusters cluster1,   cluster2  , cluster1 ,              cluster3`.  will this leads to an  unexpected behavior? 
I also checked the source code of set, https://github.com/kubernetes/apimachinery/blob/v0.22.1/pkg/util/sets/string.go#L30 https://github.com/kubernetes/apimachinery/blob/v0.22.1/pkg/util/sets/string.go#L49 seems that it has no duplicate element check.

Or the cobra framework already did this check for us?

qiujian16: sets is implemented by map[key]bool so you will not have duplicated keys

ycyaoxdu: > sets is implemented by map[key]bool so you will not have duplicated keys. But you certainly should trim suffix in advance

I just found that if any space exists between names, cobra will throw an error, so we don't need to trim that spaces. 
```
$ clusteradm addon enable --name test211,  test3 ,test1 --cluster cluster1
Deploying test211 add-on to namespaces open-cluster-management-agent-addon of managed cluster: cluster1.
Error: name is required
```

pkg/cmd/addon/disable/exec.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
+	klog.V(1).InfoS("disable options:", "dry-run", o.ClusteradmFlags.DryRun, "names", o.names, "clusters", o.clusters)
+
+	return nil
+}
+
+func (o *Options) validate() (err error) {
+	if o.names == nil {
```
qiujian16, line 29: len(o.names) == 0

pkg/cmd/addon/disable/exec.go
```diff
@@ -0,0 +1,100 @@
...
+		return fmt.Errorf("names is missing")
+	}
+
+	if o.clusters == nil {
```
qiujian16, line 33: ditto

PR #103, opened 2022-01-05
Feat: clusteradm get addon command for list enabled addon

ycyaoxdu:
Signed-off-by: ycyaoxdu <yaoyuchen0626@163.com>

Add a subcommand `list` to command `addon` for enhance the new feature 2 which mentioned in #87

PR #104, opened 2022-01-05
[WIP] Feat: clusteradm addon listall command for list all enabled addon

ycyaoxdu:
Signed-off-by: ycyaoxdu yaoyuchen0626@163.com

Add a subcommand `listall` to command `addon` for enhance the new feature 1 which mentioned in #87


itdove:
Just wondering if we can not have `list` instead of `listall`.
`clusteradm addon list` to list all enable addon on a hub.
`clusteradm addon list <managedcluster_name>` to list all enable addon on a given managedcluster
`clusteradm addon list -A` to list all enable addon on a hub and all managedclusters.

mikeshng:
I agree with @itdove `listall` doesn't seem to be a traditional param for command line tools.

itdove:
Now thinking further... I would use `get`

`clusteradm get addon` for addon on hub
`clusteradm get addon <managedcluster` for addon on managedcluster
`clusteradm get addon -A`

qiujian16:
I am thinking  whether it should be `clusteradm addon get` or `clusteradm get addon`. 

I think get is a more common command while enable/disable is specific for addon, we could have a set of command like

```
clusteradm get addon
clusteradm addon enable
clusteradm addon disable
```

and for get, we should also have
```
clusteradm get cluster
clusteradm get clusterset
clusteradm get placement
clusteradm get placemenbinding
```
I would also like to have create command for placement/clusterset  and bind command for placementbinding

ycyaoxdu:
so we will combine the 2 commands, `list` and `listall`, then refactor it from `addon list` to `get addon`, and the command `addon enable` and `addon disable` will keep as it is ?

itdove:
@qiujian16 i would prefer to keep 'clusteradm verb what' format and so have 'clusteradm enable address' but uptoyou,
I don't think 'clusteradm get placement' is necessary as it is the same as 'kubectl get placement', right? Same for others.
Cm already implement'cm get cluster' to list the managedcluster.
Also we have to be careful as cm import clusteradm (I didn't update for awhile) to not have clashes between the two.

ycyaoxdu:
so changes in this pr will be merged into #103 , close this one.

PR #107, opened 2022-01-20
Use the endpoint from --hub-apiserver for joining cluster

yue9944882:


pkg/cmd/join/exec.go
```diff
@@ -340,21 +339,10 @@ func (o *Options) createKubeConfig(externalClientUnSecure *kubernetes.Clientset,
 		return "", err
 	}
 
-	hubAPIServerInternal, err := helpers.GetAPIServer(externalClientUnSecure)
```
qiujian16, line 343: should we check if o.hubAPIServer is not set, read hubAPIServerInternal?

qiujian16 (commented):
/approve
/lgtm

Issue #84, opened 2021-12-10
`clusteradm accept` does't find the pending CSR for the applying cluster


yue9944882:
```shell
$ clusteradm accept --clusters cluster1
no CSR to approve for cluster
```
the clusteradm tools cannot find the corresponding CSR resource if the bootstrap token is signed manually instead of reusing 
 the generated tokens from `clusteradm init` commands.
 
 in our case the bootstrap token is copied from the `default` namespace instead of `open-cluster-management` which leads to the clusteradm to ignore the pending CSR from the agent that mounts the manual bootstrap token.
 
https://github.com/open-cluster-management-io/clusteradm/blob/1b0e9b887df77c11a07165f8a05a1079023e902c/pkg/cmd/accept/exec.go#L127-L128

kind of curious why are we filtering the CSRs by the `.spec.groups`?

yue9944882:
cc @qiujian16

yue9944882:
cc @champly

qiujian16:
I think we can add a flag to skip the csr check.

mgold1234:
fixed
/close

mgold1234:
fixed
/close

PR #111, opened 2022-01-24
Refactor: add e2e test using ginkgo framework

ycyaoxdu:

opened for issue: https://github.com/open-cluster-management-io/community/issues/118 , have refactored the e2e(functional) test using ginkgo framework.


test/functional/clusteradm/clusteradmE2E_test.go
```diff
@@ -0,0 +1,252 @@
...
+	"testing"
+
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+
+	"open-cluster-management.io/clusteradm/test/functional/util"
+)
+
+var e2e, e2e2, e2e3 *util.TestE2eConfig
```
zhiweiyin318, line 16: it is better to define the var in each case

test/functional/clusteradm/clusteradmE2E_test.go
```diff
@@ -0,0 +1,252 @@
...
+		// clean
+		e2e2.KindDeleteCluster()
+	})
+
+	ginkgo.Context("join hub scenario", func() {
+		ginkgo.By("join hub with bootstrap token")
+
+		ginkgo.It("should managedcluster join hub successfully", func() {
+			ginkgo.By("clusteradm version check")
+			err := e2e2.RunCmd(e2e2.Streams, "version")
```
zhiweiyin318, line 139: e2e2.RunCmd("version")
Streams is the filed of e2e2, not need as input parameter.
use the name clusteradm("version")  may be better understand.

test/functional/clusteradm/clusteradmE2E_test.go
ycyaoxdu: ok

test/functional/clusteradm/clusteradmE2E_test.go
ycyaoxdu: The output will be redirected in some case, such as this: :https://github.com/open-cluster-management-io/clusteradm/pull/111/files#diff-d092bd45baec86399bbd7cf1c6a927c365a7db10ce559b3aca73e35604fce025R73 .
So i added the stream parameter. 
And I'm wondering whether i should remove this stream parameter and then define another function to handle these cases, or keep this parameter as it is?

And for the function name, i think your suggestion is really good, i will rename that function.

ycyaoxdu:
#84 occurs

test/functional/clusteradm/clusteradmE2E_test.go
```diff
@@ -0,0 +1,266 @@
...
+
+	"open-cluster-management.io/clusteradm/test/functional/util"
+)
+
+func TestE2EClusteradm(t *testing.T) {
+	gomega.RegisterFailHandler(ginkgo.Fail)
+	ginkgo.RunSpecs(t, "E2E clusteradm test")
+}
+
+var _ = ginkgo.Describe("test clusteradm with service account", func() {
```
zhiweiyin318, line 18: suggest  use different files to include different cases.
you can refer https://github.com/open-cluster-management-io/registration-operator/tree/main/test/e2e
usually we have a e2e entry file named [e2e_suite_test.go](https://github.com/open-cluster-management-io/registration-operator/blob/main/test/e2e/e2e_suite_test.go)

test/functional/clusteradm/clusteradmE2E_test.go
```diff
@@ -0,0 +1,266 @@
...
+		ginkgo.By("join hub with service account")
+
+		ginkgo.It("should managedcluster join hub successfully", func() {
+			ginkgo.By("clusteradm version check")
+			err := e2e.Clusteradm().Version().Run()
+			gomega.Expect(err).NotTo(gomega.HaveOccurred(), "clusteradm version check error")
+
+			ginkgo.By("init hub")
+			jn, err := e2e.Clusteradm().Init(
+				"--context", e2e.Cluster().Hub().Context(),
```
zhiweiyin318, line 48: the kind clusters are created before run e2e, right? 
could we set the info of clusters as a variables, make it easy to understand, or add more comments?

test/e2e/clusteradm/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+---
+## Cluster()
+You can get cluster informations by call Cluster(). Such as:
+```
+// name of hub
+e2e.Cluster().Hub().Name()
+
+// context of managedcluster1
+e2e.Cluster().ManagedCluster1().Context()
```
qiujian16, line 24: Is this based on the assumption that all tests are run on kind? Is it possible that we have another kuberentes (non-kind) to run e2e. It should be also doc'ed here.

ycyaoxdu: yes, assuming that we use kind to start clusters, i will add comments to explain it.

ycyaoxdu:
add non-kind clusters support, user needs to specify the cluster create command format before setup, https://github.com/open-cluster-management-io/clusteradm/pull/111/files#diff-8a2c59005a01c6a3ffdca09c00a682adbebf7800f6bb6551391260815569133fR18-R23 
@qiujian16 now user can determine what kubernetes cluster they use.

ycyaoxdu:
RESTConfig  seems redundant, removed.

ycyaoxdu:
rebased

test/e2e/clusteradm/README.md
```diff
@@ -0,0 +1,127 @@
...
+`Clean(args ...string)` ,
+`Install(args ...string)` ,
+`Proxy(args ...string)` ,
+`Unjoin(args ...string)`.
+
+---
+### reset e2e environment
+
+#### 1. ResetEnv
+First, we define an intial e2e environment state: 1 hub, 2 managed(cluster1, cluster2), hub is intialized, cluster1 is join and accepted, cluster2 is just created.
```
qiujian16, line 55: initial

test/e2e/clusteradm/README.md
```diff
@@ -0,0 +1,117 @@
...
+---
+
+
+## write test
+
+Now we will show you how to write tests for clusteradm command.
+
+0.  Before we start, you should figure out what cluseter are you using, and the corresponding `create cluster` command format, `delete cluster` command format and `cluster context` format, which you should use `<cluster-name>` to replace the real cluster name. See the following example:
+
+    If we're using [kind](), the create cluster command format would be `kind crea  te cluster --name <cluster-name>`, the delete cluster command format would be `kind delete cluster --name <cluster-name>`, the context format would be `kind-<cluster-name>`.
```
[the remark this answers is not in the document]
qiujian16: is [kind]() correct?

qiujian16 (commented):
thanks, I think it makes more sense now, Just some nit. It looks good in general

/approve

test/e2e/clusteradm/README.md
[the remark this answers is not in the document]
ycyaoxdu: actually, this part is departed, sorry for forget to modify this paragraph. 
 and i have rewrote the readme

ycyaoxdu:
/hold 
@qiujian16 
I think `cmd.run()` is still not easy to understand. 
So I'm trying to remove `run()/output()` method and save the `host` and `token` in a struct member, rather than a return value, then we can use these values by calling a method looks like `e2e.xxxx.Host()`.

ycyaoxdu:
> @qiujian16 I think `cmd.run()` is still not easy to understand. So I'm trying to remove `run()/output()` method and save the `host` and `token` in a struct member, rather than a return value, then we can use these values by calling a method looks like `e2e.xxxx.Host()`.

/cc @qiujian16  done. 
1. Changed the command interface. 
2. migrate test framework from `ginkgo/v1` to `ginkgo/v2` (only for e2e test).
Please help to review and give more suggestions. thanks.

PR #146, opened 2022-03-08
Remove deploy-agent option from subscription controller

philipwu08:
The deploy-agent option was removed from the subscription controller.

https://github.com/open-cluster-management-io/multicloud-operators-subscription/issues/106

Signed-off-by: Philip Wu <phwu@redhat.com>


mikeshng (approved):
/lgtm

/approve

I also tested this change locally and it was fine.

PR #152, opened 2022-03-14
Feat: Proxy health supports selecting clusters via -c or --cluster

yue9944882:


pkg/cmd/proxy/health/exec.go
```diff
@@ -150,8 +150,10 @@ func (o *Options) run(streams genericclioptions.IOStreams) error {
 
 	w := newWriter(streams)
 	for _, cluster := range managedClusterList.Items {
-		if err := o.visit(&w, hubRestConfig, addonClient, tunnel.DialContext, cluster.Name); err != nil {
-			klog.Errorf("An error occurred when requesting: %v", err)
+		if len(o.clusters) == 0 || contains(o.clusters, cluster.Name) {
```
qiujian16, line 153: how about putting o.clusters to a sets.String?

yue9944882: updated

pkg/cmd/proxy/health/exec.go
```diff
@@ -329,3 +333,12 @@ func (w *writer) print(clusterName, installed, available, health, latency string
 func (w *writer) flush() {
 	_ = w.w.Flush()
 }
+
+func contains(vs []string, target string) bool {
```
qiujian16, line 337: this func is not needed, right?

yue9944882: nice catch

Issue #133, opened 2022-02-21
[Umbrella Issue] CLI usability improvements


yue9944882:
- __Comprehensive `clusteradm get ..` command__:
  - [x] get clusters: 
    - https://github.com/open-cluster-management-io/clusteradm/pull/124
  - [x] get addons: 
    - https://github.com/open-cluster-management-io/clusteradm/pull/103
    - https://github.com/open-cluster-management-io/clusteradm/pull/129
  - [x] get works:
    - https://github.com/open-cluster-management-io/clusteradm/pull/131
  - [x] get hub-info: 
    - https://github.com/open-cluster-management-io/clusteradm/pull/128
  - [ ] get klusterlet-info: 
  - [x] get clustersets:
    - https://github.com/open-cluster-management-io/clusteradm/pull/124
  - [ ] get placements:
  - [ ] general `-o [*table|tree|yaml]` printers
  
- __Clusterset operatibility__:
   - [x] create: https://github.com/open-cluster-management-io/clusteradm/pull/123
   - [x] add cluster/bind namespace: https://github.com/open-cluster-management-io/clusteradm/pull/126
   - [ ] remove cluster
   - [ ] unbind
 
- __Version mgmt / Upgrading__:
  - [x] https://github.com/open-cluster-management-io/clusteradm/pull/119
  
- __Proxy / tunneling__:
  - [ ] proxy health: https://github.com/open-cluster-management-io/clusteradm/pull/90
  - [ ] proxy tunnel: (creating tunnels for kubectl-access to the managed clusters)
    - create
    - connect
    - delete
   - [ ] proxy stats: (collecting stats such as in-use bandwidth, connection counts, etc.)

- __Addon management__:
  - [x] enable/disable: https://github.com/open-cluster-management-io/clusteradm/pull/101
  
- __Miscs__:
  - [ ] unjoin cluster: (removing registered managed clusters from the hub neatly)

qiujian16:
maybe also add create/delete work command?

ycyaoxdu:
now we use command `clusteradm clusterset set` to move one cluster from clusterset1 to  clusterset2 #153

ycyaoxdu:
added command `clusteradm delete clusterset` #154

PR #158, opened 2022-03-16
feat: add command clusteradm get placements

ycyaoxdu:

### DIFF
--- a/pkg/cmd/get/placement/exec.go
+++ b/pkg/cmd/get/placement/exec.go
@@ -1,0 +1,195 @@
+// Copyright Contributors to the Open Cluster Management project
+package placement
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/fatih/color"
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	clusterv1beta1 "open-cluster-management.io/api/client/cluster/clientset/versioned/typed/cluster/v1beta1"
+	"open-cluster-management.io/api/cluster/v1beta1"
+	"open-cluster-management.io/clusteradm/pkg/helpers/printer"
+)
+
+const placementLabel = "cluster.open-cluster-management.io/placement"
+
+func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
+	o.printer.Competele()
+
+	return nil
+}
+
+func (o *Options) validate(args []string) (err error) {
+	if len(args) > 1 {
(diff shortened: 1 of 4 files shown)


ycyaoxdu:
example output:
```
root@yuchen clusteradm [feat] $ clusteradm get placements
NAME                   STATUS   REASON                  SELETEDCLUSTERS
placement-policy-pod   True     AllDecisionsScheduled   1
```

ycyaoxdu:
#133  #163

qiujian16:
we might want to show more details, such as the clusters in the related decisions

ycyaoxdu:
use a slice of cluster names to replace the number
```
NAME                               STATUS       REASON                            SELETEDCLUSTERS
placement-policy-pod   True              AllDecisionsScheduled    [cluster1]
```

ycyaoxdu:
```
➜  clusteradm git:(feat/get-placement) ✗ clusteradm get placements
<Placement>
└── placement1
    └── <Namespace>: default
    └── <ClusterSet>: [default]
    └── <Status>
    │   ├── <Conditions>
    │   │   ├── PlacementSatisfied -> true
    │   │   ├── PlacementMisconfigured -> unknown
    │   ├── <Number of Selected Clusters>: 2
    └── <PlacementDecision>
        └── cluster1
        └── cluster2
➜  clusteradm git:(feat/get-placement) ✗ clusteradm get placements -o table
NAMESPACE   NAME         STATUS   REASON                  SELETEDCLUSTERS
default     placement1   True     AllDecisionsScheduled   [cluster1 cluster2]
```

@qiujian16 PTAL

qiujian16:
should we print table by default or tree by default? I think we are not quite consistent on other commands.

ycyaoxdu:
> should we print table by default or tree by default? I think we are not quite consistent on other commands.

@qiujian16  maybe we should use tree as the default get output format? some command can not print as table, like `hub-info`, `klusterlet-info` , `token`.

qiujian16:
some may not have tree format, like `get clusters`? Or we should also change get clusters default format

ycyaoxdu:
> should we print table by default or tree by default? I think we are not quite consistent on other commands.

currently i've set tree to default. i will

> 

ok, will do this in https://github.com/open-cluster-management-io/clusteradm/pull/174

pkg/cmd/get/placement/exec.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+const placementLabel = "cluster.open-cluster-management.io/placement"
+
+func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
+	return nil
+}
+
+func (o *Options) validate(args []string) (err error) {
+	if len(args) > 0 {
```
qiujian16, line 28: nit, we should allow input name of the placement. And then you could use list with fieldSelector.

PR #161, opened 2022-03-18
modify output while klusterlet not found

ycyaoxdu:


qiujian16 (commented):
/approve
/lgtm

Issue #162, opened 2022-03-18
bug: command create work cannot create correct manifest work


ycyaoxdu:
deploy in managedcluster create failed while i create a work using clusteradm 
```
status:
  conditions:
  - lastTransitionTime: "2022-03-18T03:28:21Z"
    message: 1 of 1 resources have unknown status
    observedGeneration: 1
    reason: ResourcesStatusUnknown
    status: Unknown
    type: Available
  - lastTransitionTime: "2022-03-18T03:28:21Z"
    message: Failed to apply manifest work
    observedGeneration: 1
    reason: AppliedManifestWorkFailed
    status: "False"
    type: Applied
  resourceStatus:
    manifests:
    - conditions:
      - lastTransitionTime: "2022-03-18T03:28:21Z"
        message: 'Failed to apply manifest: cluster1 "hello-work-demo" is forbidden:
          User "system:serviceaccount:open-cluster-management-agent:klusterlet-work-sa"
          cannot get resource "cluster1" in API group "" at the cluster scope'
        reason: AppliedManifestFailed
        status: "False"
        type: Applied
      - lastTransitionTime: "2022-03-18T03:28:21Z"
        message: Resource meta is incompleted
        reason: IncompletedResourceMeta
        status: Unknown
        type: Available
      resourceMeta:
        group: work.open-cluster-management.io
        kind: ManifestWork
        name: hello-work-demo
        namespace: cluster1
        ordinal: 0
        resource: ""
        version: v1
      statusFeedback: {}
```

qiujian16:
we might need to better document here. The manifest is not to be `ManifestWork`, but a manifest files containing legacy kube resources.

ycyaoxdu:
the spec should be:
```
spec:
  workload:
    manifests:
    - apiVersion: apps/v1
      kind: Deployment
      metadata:
        name: hello
        namespace: default
      spec:
        selector:
          matchLabels:
            app: hello
        template:
          metadata:
            labels:
              app: hello
          spec:
            containers:
            - command:
              - sh
              - -c
              - echo "Hello, Kubernetes!" && sleep 3600
              image: quay.io/asmacdo/busybox
              name: hello
```

but while we use clusteradm to create, it would be:
```
spec:
  workload:
    manifests:
    - apiVersion: work.open-cluster-management.io/v1
      kind: ManifestWork
      metadata:
        name: hello-work-demo
        namespace: cluster1
      spec:
        workload:
          manifests:
[...]

PR #164, opened 2022-03-21
add long description to create work command

ycyaoxdu:

for https://github.com/open-cluster-management-io/clusteradm/issues/162#issuecomment-1073411675

### DIFF
--- a/pkg/cmd/create/work/cmd.go
+++ b/pkg/cmd/create/work/cmd.go
@@ -22,8 +22,8 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 
 	cmd := &cobra.Command{
 		Use:          "work",
-		Short:        "create a work",
-		Long:         "create manifestwork on a specified managedcluster",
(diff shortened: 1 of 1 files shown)


pkg/cmd/create/work/cmd.go
```diff
@@ -22,8 +22,8 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 
 	cmd := &cobra.Command{
 		Use:          "work",
-		Short:        "create a work",
-		Long:         "create manifestwork on a specified managedcluster",
+		Short:        "create a work using resource-to-apply yaml file",
+		Long:         "create a work using resource-to-apply yaml file. The resource-to-apply yaml file is exactly the yaml file you want to apply on the managed cluster. Take care and do not use a ManifestWork yaml file here.",
```
qiujian16, line 26: using a file containing common kubernetes resource manifests, or a director containing a set of manifest files. 

We do not need to say don't use ManifestWork file, it is a bit confusing.

Issue #159, opened 2022-03-18
bug: delete work cannot finish executing


ycyaoxdu:
Ran `clusteradm delete work addon-work-manager-deploy  --cluster itdove2-4-10` and it never finished, maybe because the cluster is already removed.

```
 oc get managedcluster 
NAME           HUB ACCEPTED   MANAGED CLUSTER URLS                                                      JOINED   AVAILABLE   AGE
itdove-4-10    true           https://api.aws-4-10-rc4-sno-2x-vxxxxx:6443/   True     True        16d
itdove2-4-10   true           https://api.aws-4-10-rc4-sno-2x-xxxxxx:6443/   True     Unknown     13d
```

When the cluster is "available" it works.
reported by @itdove

PR #165, opened 2022-03-21
fix: set timeout for delete work

ycyaoxdu:

#159 set a timeout for deleting work.

### DIFF
--- a/pkg/cmd/delete/work/exec.go
+++ b/pkg/cmd/delete/work/exec.go
@@ -4,6 +4,7 @@ package work
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"k8s.io/apimachinery/pkg/watch"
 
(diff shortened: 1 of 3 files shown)


qiujian16 (commented):
I think we should also add a `--force` flag so we delete the work at first, and if it is not removed, remove its finalizer to force remove it.

ycyaoxdu:
```
root@yuchen clusteradm [fix] $ clusteradm get clusters
NAME       ACCEPTED   AVAILABLE   CLUSTERSET   CPU   MEMORY       KUBERENETES VERSION
cluster2   true       Unknown                  8     16185152Ki   v1.21.1
root@yuchen clusteradm [fix] $ clusteradm get works --cluster cluster2
NAME    CLUSTER    NUMBER OF MANIFESTS   APPLIED   AVAILABLE
work2   cluster2   1                     True      True
root@yuchen clusteradm [fix] $ clusteradm delete work work2 --cluster cluster2 
Error: delete work work2 timeout, failed to delete
root@yuchen clusteradm [fix] $ clusteradm delete work work2 --cluster cluster2 --force
work work2 in cluster cluster2 is deleted
root@yuchen clusteradm [fix] $ clusteradm get works --cluster cluster2

```

pkg/cmd/delete/work/exec.go
```diff
@@ -51,21 +52,53 @@ func (o *Options) run() error {
...
+
+	if o.Force {
+		work.ObjectMeta.Finalizers = work.ObjectMeta.Finalizers[:0]
+		_, err = workClient.WorkV1().ManifestWorks(o.Cluster).Update(context.TODO(), work, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+	}
+
+	err = workClient.WorkV1().ManifestWorks(o.Cluster).Delete(context.TODO(), o.Workname, metav1.DeleteOptions{})
```
qiujian16, line 94: run delete at first then remove finalizer, otherwise there is chance that finalizer is added back

ycyaoxdu: done

pkg/cmd/delete/work/exec.go
```diff
@@ -51,21 +52,53 @@ func (o *Options) run() error {
...
+
+	}(errChannel)
+
+	err = workClient.WorkV1().ManifestWorks(o.Cluster).Delete(context.TODO(), o.Workname, metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	if o.Force {
+		work.ObjectMeta.Finalizers = work.ObjectMeta.Finalizers[:0]
```
qiujian16, line 92: nit you probably want to get again here.

pkg/cmd/delete/work/exec.go
```diff
@@ -51,21 +52,64 @@ func (o *Options) run() error {
...
+
+	err = workClient.WorkV1().ManifestWorks(o.Cluster).Delete(context.TODO(), o.Workname, metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+
+	if o.Force {
+		// check whether work is already deleted, if not, remove the finalizer
+		work, err = workClient.WorkV1().ManifestWorks(o.Cluster).Get(context.TODO(), o.Workname, metav1.GetOptions{})
+		if err != nil {
```
qiujian16, line 94: I would do 

```
if errors.IsNotFound(err) {
	fmt.Fprintf(o.Streams.Out, "work %s is deleted\n", o.Workname)
	return nil
}

if err != nil {
    return err
}

if len(work.Finalizers) == 0 {
    return nil
}

```

qiujian16 (commented):
/approve
/lgtm

PR #167, opened 2022-03-21
add support for manual bootstrap token for accept command.

mgold1234:
related issue - #84

### DIFF
--- a/pkg/cmd/accept/exec.go
+++ b/pkg/cmd/accept/exec.go
@@ -117,31 +117,39 @@ func (o *Options) approveCSR(kubeClient *kubernetes.Clientset, clusterName strin
 		return false, err
 	}
 	var csr *certificatesv1.CertificateSigningRequest
-	for _, item := range csrs.Items {
-		//Does not have the correct name prefix
-		if !strings.HasPrefix(item.Spec.Username, userNameSignatureBootstrapPrefix) &&
-			!strings.HasPrefix(item.Spec.Username, userNameSignatureSA) {
-			continue
+	var passedCSRs []certificatesv1.CertificateSigningRequest
+	if o.skipApproveCheck {
+		passedCSRs = csrs.Items
+	} else {
+		for _, item := range csrs.Items {
+			//Does not have the correct name prefix
+			if !strings.HasPrefix(item.Spec.Username, userNameSignatureBootstrapPrefix) &&
+				!strings.HasPrefix(item.Spec.Username, userNameSignatureSA) {
+				continue
+			}
+			//Check groups
+			groups := sets.NewString(item.Spec.Groups...)
+			if !groups.Has(groupNameBootstrap) &&
+				!groups.Has(groupNameSA) {
+				continue
+			}
+			passedCSRs = append(passedCSRs, item)
 		}
-		//Check groups
-		groups := sets.NewString(item.Spec.Groups...)
-		if !groups.Has(groupNameBootstrap) &&
-			!groups.Has(groupNameSA) {
-			continue
-		}
+	}
+	for _, passedCSR := range passedCSRs {
 		//Check if already approved or denied
-		approved, denied := GetCertApprovalCondition(&item.Status)
-		//if alreaady denied, then nothing to do
+		approved, denied := GetCertApprovalCondition(&passedCSR.Status)
+		//if already denied, then nothing to do
 		if denied {
-			fmt.Printf("CSR %s already denied\n", item.Name)
+			fmt.Printf("CSR %s already denied\n", passedCSR.Name)
 			return true, nil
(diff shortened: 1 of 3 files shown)


pkg/cmd/accept/exec.go
```diff
@@ -24,11 +24,13 @@ import (
 const (
 	groupNameBootstrap               = "system:bootstrappers:managedcluster"
 	userNameSignatureBootstrapPrefix = "system:bootstrap:"
-	userNameSignatureSA              = "system:serviceaccount:open-cluster-management:cluster-bootstrap"
+	userNameSignatureSAPrefix         = "system:serviceaccount:"
```
qiujian16, line 27: I do not think we should change the sa name here. It is to ensure that only specifit sa can be approved

pkg/cmd/accept/exec.go
```diff
@@ -24,11 +24,13 @@ import (
...
 	groupNameSA                      = "system:serviceaccounts:open-cluster-management"
+	groupNameSAManualToken           = "system:serviceaccounts:default"
```
qiujian16, line 29: this means we are trying an sa in default ns, I do not think it is a common case.

qiujian16 (commented):
thanks. I would suggest an alternative approach. Maybe we could introduce a flag `--skip-approve-check`, if this flag is set, we do not check the prefix or requester in the csr, but approve the csr directly.

pkg/cmd/accept/exec.go
```diff
@@ -118,16 +118,18 @@ func (o *Options) approveCSR(kubeClient *kubernetes.Clientset, clusterName strin
...
-		if !strings.HasPrefix(item.Spec.Username, userNameSignatureBootstrapPrefix) &&
-			!strings.HasPrefix(item.Spec.Username, userNameSignatureSA) {
-			continue
-		}
-		//Check groups
-		groups := sets.NewString(item.Spec.Groups...)
-		if !groups.Has(groupNameBootstrap) &&
-			!groups.Has(groupNameSA) {
-			continue
+		if !o.skipApproveCheck {
```
qiujian16, line 121: /approve

hrm, I think we need some refactor here. It seems we only approve only csr in the list, but we probably need to approve them all. Would you add a TODO before line 121 on we need to approve all CSRs that meets the check, and if the check is skipped, just approve all csrs in this list?

pkg/cmd/accept/exec.go
```diff
@@ -117,18 +117,22 @@ func (o *Options) approveCSR(kubeClient *kubernetes.Clientset, clusterName strin
...
-		if !strings.HasPrefix(item.Spec.Username, userNameSignatureBootstrapPrefix) &&
-			!strings.HasPrefix(item.Spec.Username, userNameSignatureSA) {
-			continue
-		}
-		//Check groups
-		groups := sets.NewString(item.Spec.Groups...)
-		if !groups.Has(groupNameBootstrap) &&
-			!groups.Has(groupNameSA) {
-			continue
+	if !o.skipApproveCheck {
```
qiujian16, line 120: you probably want some thing like

```go
passedCSRs := []certificatesv1.CertificateSigningRequest
if o.skipApproveCheck {
  passedCSRs = csrs.Items
} else {
  for _, item := range csrs.Items {
    // check name and group
    passedCSRs = append(passedCSRs, item)
  }
}

// check whether they are approved in passedCSRs

```

pkg/cmd/accept/exec.go
```diff
@@ -117,31 +117,42 @@ func (o *Options) approveCSR(kubeClient *kubernetes.Clientset, clusterName strin
...
+					continue
+				}
+			}
 		}
+	}
+	for _, passedCSR :=range passedCSRs{
 		//Check if already approved or denied
-		approved, denied := GetCertApprovalCondition(&item.Status)
-		//if alreaady denied, then nothing to do
+	approved, denied := GetCertApprovalCondition(&passedCSR.Status)
```
qiujian16, line 144: nit, lint issue?

pkg/cmd/accept/exec.go
```diff
@@ -117,31 +117,42 @@ func (o *Options) approveCSR(kubeClient *kubernetes.Clientset, clusterName strin
...
-		if !groups.Has(groupNameBootstrap) &&
-			!groups.Has(groupNameSA) {
-			continue
+	var passedCSRs []certificatesv1.CertificateSigningRequest
+	if o.skipApproveCheck {
+		passedCSRs = csrs.Items
+	} else {
+		for _, item := range csrs.Items {
+			// check name and group
+			passedCSRs = append(passedCSRs, item)
```
qiujian16, line 126: I think you should put it after `if`, right?

PR #170, opened 2022-03-22
Add policy-framework to hub-addon install

dhaiducek:
Adds `clusteradm` command for `policy-framework`:

```bash
clusteradm install hub-addon --name policy-framework
```

I'm debating how these should be named and whether we should have different names between `clusteradm install hub-addon` and `clusteradm addon enable` or use 'policy-framework' for both core components (this would match app's current behaviors). The `config-policy-controller` could remain a separate name for `clusteradm addon enable` or we could roll it in...

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,7 +51,7 @@ clusteradm <cmd> [subcmd] [flags]
 
 ## Client
 
-- The [main](cmd/clusteradm.go) provides a cmdutil.Factory which can be laverage to get different clients and also the *rest.Config. The factory can be passed to the cobra.Command and then save in the Options.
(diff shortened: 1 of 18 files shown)


JustinKuli (approved):

dhaiducek:
>I wish the first command was more like `clusteradm install hub-addon`

Yes! I couldn't quite put my finger on it, but I agree that this is the best solution.

>If/when https://github.com/open-cluster-management-io/enhancements/pull/44 is accepted and implemented, would this command be updated to deploy the policy-addon-controller as well?

Correct.

>In addition, I think we would implement addon enable commands to deploy the policy components to managed clusters, is that right? Just checking I understand the intended plan.

Yes.

PR #171, opened 2022-03-23
clusteradm get addon --addon: filter by addon name

yue9944882:


qiujian16:
should we just do 

`clusteradm get addon <addon name>`?

qiujian16 (commented):
/approve
/lgtm

PR #174, opened 2022-03-24
feat: add printer supporting tree, table and yaml format

ycyaoxdu:

### DIFF
--- a/pkg/cmd/get/cluster/cmd.go
+++ b/pkg/cmd/get/cluster/cmd.go
@@ -50,5 +50,7 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 
 	cmd.Flags().StringVar(&o.Clusterset, "clusterset", "", "ClusterSet of the clusters")
 
+	o.printer.AddFlag(cmd.Flags())
+
 	return cmd
 }
--- a/pkg/cmd/get/cluster/exec.go
+++ b/pkg/cmd/get/cluster/exec.go
@@ -11,9 +11,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	clusterclientset "open-cluster-management.io/api/client/cluster/clientset/versioned"
 	clusterapiv1 "open-cluster-management.io/api/cluster/v1"
+	"open-cluster-management.io/clusteradm/pkg/helpers/printer"
 )
 
 func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
+	o.printer.Competele()
+
 	return nil
 }
 
@@ -21,6 +24,12 @@ func (o *Options) validate(args []string) (err error) {
 	if len(args) != 0 {
 		return fmt.Errorf("there should be no argument")
 	}
+
+	err = o.printer.Validate()
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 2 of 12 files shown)


ycyaoxdu:
#133 
some command may do not need yaml output? such as `hub-info`, `klusterlet-info`? so i just added -o flag for which may need yaml output. 
i think i need more suggestions on this...

itdove:
@ycyaoxdu Can't you use something like this: https://github.com/open-cluster-management-io/clusteradm/blob/cae3fa5a804dac98894c710c4baa0eefbf902291/pkg/cmd/get/cluster/exec.go#L54 which gives you all the -o options instead only the YAML.

ycyaoxdu:
@itdove i've added a field to genericclioptions called Printer https://github.com/open-cluster-management-io/clusteradm/pull/174/commits/b4d35a6b06b569f71aa93a8e892b941a4e1ed42b
and then the usage is like https://github.com/open-cluster-management-io/clusteradm/pull/174/commits/9ad754ed79b30c6e9c58d7abf86bbe0ccc916172 
, how about this?

---

> return o.printer.PrintObj(table, o.Streams.Out)

and also used this resourcePrinter to do this.

itdove:
I don't think a genericoption is the good choice as I think `-o` doesn't apply to all commands but only on `get` commands... 
as you can see here https://github.com/stolostron/cm-cli/blob/5bbe69bf8312ef615d31d5b5037ae0c0174cb0c7/pkg/cmd/get/clusterclaim/cmd.go#L52-L54 I added the `-o` to each command I think they can use it. I didn't add it to all gets because some of the get cmd are not supposed to have a `-o` like `cm get config cluster`.

ycyaoxdu:
> I don't think a genericoption is the good choice as I think `-o` doesn't apply to all commands but only on `get` commands... as you can see here https://github.com/stolostron/cm-cli/blob/5bbe69bf8312ef615d31d5b5037ae0c0174cb0c7/pkg/cmd/get/clusterclaim/cmd.go#L52-L54 I added the `-o` to each command I think they can use it. I didn't add it to all gets because some of the get cmd are not supposed to have a `-o` like `cm get config cluster`.

hi @itdove I've noticed that we need to add [apis](https://github.com/stolostron/cm-cli/tree/5bbe69bf8312ef615d31d5b5037ae0c0174cb0c7/api/cm-cli/v1alpha1) to generalize the --output ? could you please help to contribute this feature? or give me some advice on this? thanks a lot.

ycyaoxdu:
@qiujian16 PTAL 
add a printer supporting tree, table, yaml format
```
➜  clusteradm git:(feat/general-printers) clusteradm get clusters
<ManagedCluster> 
└── <cluster1> 
│   ├── <KubernetesVersion> v1.24.0
│   ├── <Capacity> 
│   │   ├── <Cpu> 16
│   │   ├── <Memory> 32763764Ki
│   ├── <Accepted> true
│   ├── <Available> True
│   ├── <ClusterSet> default
└── <cluster2> 
    └── <Capacity> 
    │   ├── <Cpu> 16
    │   ├── <Memory> 32763764Ki
    └── <Accepted> true
    └── <Available> True
    └── <ClusterSet> default
    └── <KubernetesVersion> v1.24.0
➜  clusteradm git:(feat/general-printers) clusteradm get clustersets
<ManagedClusterSet> 
└── <default> 
│   ├── <Status> 2 ManagedClusters selected
│   ├── <BoundNamespace> default
└── <global> 
    └── <BoundNamespace> 
    └── <Status> 2 ManagedClusters selected
➜  clusteradm git:(feat/general-printers) clusteradm get works --cluster cluster1
<ManifestWork> 
└── <my-first-work> 
│   ├── <Cluster> cluster1
│   ├── <Number of Manifests> 2
│   ├── <Applied> False
│   ├── <Available> True
└── <example-manifestwork> 
    └── <Cluster> cluster1
    └── <Number of Manifests> 2
    └── <Applied> True
    └── <Available> True
```
default format has set to tree.

pkg/cmd/get/cluster/cmd.go
```diff
@@ -49,6 +49,7 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	}
 
 	cmd.Flags().StringVar(&o.Clusterset, "clusterset", "", "ClusterSet of the clusters")
+	cmd.Flags().StringVarP(&o.output, "output", "o", "tree", "output format can be tree, table or yaml")
```
qiujian16, line 52: I think it is better you have a common PrinterOption and addFlags here.

You can initialize the printer with what real tree printer or table printer wants to do here.

pkg/cmd/get/cluster/exec.go
```diff
@@ -50,8 +66,32 @@ func (o *Options) run() (err error) {
...
+	clusters.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   clusterGroup,
+		Version: clusterAPIVersion,
+		Kind:    clusterKind,
+	})
+
+	o.printer.WithTree(tree).WithTable(table).WithYaml(clusters)
 
-	return o.printer.PrintObj(table, o.Streams.Out)
+	return o.printer.Print(o.Streams)
```
qiujian16, line 78: can we have a func

```
func converToTree(obj runtime.Object, tree *printer.TreePrinter ) *printer.TreePrinter

func converToTable(obj runtime.Object) *metav1.Table
```

so when call printers, we have

```
o.printer.WithTreeConverter(convertToTree).WithTableConverter(convertToTable)
```

so when you print, you just can call

```
o.printer.Print(o.Streams, clusters)
```

qiujian16 (commented):
/approve
/lgtm

PR #175, opened 2022-03-24
Rename `addons` --> `hub-addon`

dhaiducek:
Clarify subcommand referencing "addon" since the word is overloaded in OCM.


pkg/cmd/install/hubAddon/cmd.go
```diff
@@ -1,5 +1,5 @@
 // Copyright Contributors to the Open Cluster Management project
-package addons
+package hubAddon
```
JustinKuli, line 2: Usual go convention would be to keep this all lowercase: https://go.dev/blog/package-names

dhaiducek:
Sorry for the noise--I didn't know Git was case insensitive for paths by default! Should be good now.

JustinKuli (approved):

PR #179, opened 2022-03-25
Fix output format issue of join

jichenjc:
The format of output is wrong:

```
"join options:" dry-run=true cluster="cluster1" api-server="https://127.0.0.1:35123" ="(MISSING)"
```


qiujian16 (commented):
/approve
/lgtm

PR #180, opened 2022-03-25
Add more log for showing the image version in join command

jichenjc:
output looks like this 
```
I0325 10:10:25.105009  872082 exec.go:57] 'registration image version'="v0.6.0" 'placement image version'="v0.3.0" 'work image version'="v0.6.0" 'operator image version'="v0.6.0"
```

### DIFF
--- a/pkg/cmd/join/exec.go
+++ b/pkg/cmd/join/exec.go
@@ -53,6 +53,11 @@ func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
 		WorkImageVersion:         versionBundle.Work,
 		OperatorImageVersion:     versionBundle.Operator,
 	}
+	klog.V(3).InfoS("Image version:",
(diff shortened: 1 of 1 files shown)


pkg/cmd/join/exec.go
```diff
@@ -54,6 +54,11 @@ func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
 		WorkImageVersion:         versionBundle.Work,
 		OperatorImageVersion:     versionBundle.Operator,
 	}
+	klog.V(3).InfoS(
```
yue9944882, line 57: > InfoS(msg string, keysAndValues ...interface{}) 

i think an empty string is needed as the first argument

jichenjc: right, updated just now

yue9944882 (commented):
/lgtm
/approve

PR #195, opened 2022-04-06
fix: use poll method to monitor namespace deletion

ycyaoxdu:

Since the error `unable to create new content in namespace open-cluster-management-agent because it is being terminated` always leads to panic, use poll to monitor resources to be terminated.

### DIFF
--- a/test/e2e/util/util.go
+++ b/test/e2e/util/util.go
@@ -4,7 +4,8 @@ package util
 import (
 	"fmt"
 	"os"
-	"time"
+
+	"open-cluster-management.io/clusteradm/pkg/config"
 )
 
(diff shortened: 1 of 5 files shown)


test/e2e/util/util.go
```diff
@@ -46,8 +46,8 @@ func initE2E() *TestE2eConfig {
 		e2eConf.clusteradm.h = HandledOutput{}
 
 		// wait for resources to terminate
-		t := 30
-		fmt.Fprintf(os.Stdout, "wait resources to be terminated for %d seconds", t)
+		t := 60
```
qiujian16, line 49: it is better that we poll on whether ns still exists?

ycyaoxdu: done

PR #200, opened 2022-04-07
Add WithCache and getCache

itdove:

### DIFF
--- a/pkg/helpers/apply/applierbuilder.go
+++ b/pkg/helpers/apply/applierbuilder.go
@@ -41,13 +41,22 @@ type iApplierBuilder interface {
 	WithTemplateFuncMap(fm template.FuncMap) *ApplierBuilder
 	//WithOwner add an ownerref to the object
 	WithOwner(owner runtime.Object, blockOwnerDeletion, controller bool, scheme *runtime.Scheme) *ApplierBuilder
+	//WithCache add cache
+	WithCache(cache resourceapply.ResourceCache) *ApplierBuilder
 }
 
-var _ iApplierBuilder = &ApplierBuilder{}
+var _ iApplierBuilder = NewApplierBuilder()
+
+//New ApplyBuilder
+func NewApplierBuilder() *ApplierBuilder {
+	return &ApplierBuilder{}
+}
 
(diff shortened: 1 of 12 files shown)


pkg/helpers/apply/applierbuilder.go
```diff
@@ -76,3 +85,12 @@ func (a *ApplierBuilder) WithOwner(owner runtime.Object, blockOwnerDeletion, con
 	a.scheme = scheme
 	return a
 }
+
+func (a *ApplierBuilder) WithCache(cache resourceapply.ResourceCache) *ApplierBuilder {
+	a.cache = NewResourceCache()
```
qiujian16, line 90: why newCache in withCache, given you have a cache as input of the func.

pkg/helpers/apply/applierbuilder.go
```diff
@@ -76,3 +85,12 @@ func (a *ApplierBuilder) WithOwner(owner runtime.Object, blockOwnerDeletion, con
 	a.scheme = scheme
 	return a
 }
+
+func (a *ApplierBuilder) WithCache(cache resourceapply.ResourceCache) *ApplierBuilder {
```
qiujian16, line 89: you might want to comment in what cases this should be used.

pkg/helpers/apply/applierbuilder.go
itdove: 1. To avoid the developer to have to call the NewCache from the ressourceapply package.
2. WithCache allows the developer to use another implementation of the cache if he doesn't want to use the one provided by resourceappply.

pkg/helpers/apply/applierbuilder.go
itdove: Ok

pkg/helpers/apply/applierbuilder.go
qiujian16: so that means in this func you should do like `a.cache = cache`, right?

itdove: oops

qiujian16 (commented):
/lgtm

Issue #122, opened 2022-02-09
Unexported struct functions for Option struct


prometherion:
I'm eager for using the `clusteradm` repository as a package for my automation tools and noticed that the [Option](https://github.com/open-cluster-management-io/clusteradm/blob/ba85108480ae137716a4358aa7e9e4f23fda0271/pkg/cmd/accept/options.go#L9-L18) struct has several functions not exported, making it impossible to use outside of its package.

Would be great to understand what's the reason for closing (_sic_) this and if you're open to contributions on providing visibility to these functions. As I said, I'd like to use the commands from my code and not launch a process using the bare binary.

qiujian16:
Hi I do not think there is a strong reason to make it private. Any PR is welcomed! :)

itdove:
Correct, there is no strong reason to make some methods/variable public.

I would suggest to make public:
1. The Option struct and all variables in that sturct.
2. The NewOption method
3. The Validate and Run method.
I don't think the newCmd nor the complete should be make public as they are tight to the cobra framework.
Also maybe some tests done in the cobra PreRunE could be moved to the validate (optional).
@qiujian16 thought?
@mgold1234 FYI

ycyaoxdu:
@prometherion maybe you can close this issue now? thanks :)

PR #206, opened 2022-04-12
Make some Variable and Method to be at public scope

mgold1234:
Make some Variable and Method to be at public scope to use outside the package 

realted to issue:#122


qiujian16 (commented):
/approve
/lgtm

PR #207, opened 2022-04-13
change method and veriables to public scope in addon command.

mgold1234:
related to issue #122


yue9944882 (commented):
/lgtm

@qiujian16 how do you think about it?

PR #216, opened 2022-04-24
change method and veriables to public scope in clean command.

mgold1234:
related issue #122


mgold1234:
@qiujian16 , can you please  cr

PR #224, opened 2022-05-03
Stop using the GitHub API in the install script

mprahl:
The install script provides a strange error when the GitHub API rate
limit is encountered. This commit removes the need to use the GitHub
API and bypasses rate limiting checks.


yue9944882 (commented):
/lgtm
/approve

thanks!

PR #229, opened 2022-05-24
Add the `governance-policy-addon-controller`

dhaiducek:
Addresses:
- https://github.com/stolostron/backlog/issues/16046
- https://github.com/open-cluster-management-io/community/issues/122


pkg/cmd/install/hubaddon/scenario/addon/policy/addon-controller_clusterrole.yaml
```diff
@@ -0,0 +1,222 @@
...
+  - get
+  - list
+  - watch
+- apiGroups:
+  - addon.open-cluster-management.io
+  resourceNames:
+  - cert-policy-controller
+  - config-policy-controller
+  - governance-policy-framework
+  - iam-policy-controller
```
JustinKuli, line 23: Doesn't need cert- or iam- (multiple times in this file)

pkg/cmd/install/hubaddon/scenario/addon/policy/addon-controller_configmap.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright Contributors to the Open Cluster Management project
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    app: governance-policy-addon-controller
+  name: policy-addon-ctrl-manager-config
+  namespace: open-cluster-management
+data:
+  controller_manager_config.yaml: |
```
JustinKuli, line 10: I don't think we use this configmap - I think we would need to make changes in the addon-controller to accept configuration like this, as opposed to (or in addition to) the command line flags. See https://book.kubebuilder.io/component-config-tutorial/tutorial.html

JustinKuli (changes requested):
Mostly looks good! Just a few small changes.

One other thought: is changing `name` to `names` and `cluster` to `clusters` required in this PR? Or would it make sense in a separate PR so it can be discussed? I like the change, but we might need feedback on whether this "breaks" an API or anything

pkg/cmd/install/hubaddon/scenario/addon/policy/addon-controller_configmap.yaml
dhaiducek: Deleting...

dhaiducek:
/hold for flag/spelling fixes in:
- https://github.com/open-cluster-management-io/clusteradm/pull/231

mprahl (approved):
Nice!

pkg/cmd/addon/enable/cmd.go
```diff
@@ -52,6 +68,7 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	cmd.Flags().StringVarP(&o.Namespace, "namespace", "n", "open-cluster-management-agent-addon", "Specified namespace to addon addon")
 	cmd.Flags().StringSliceVar(&o.Clusters, "clusters", []string{}, "Names of the managed cluster to deploy the add-on to (comma separated)")
 	cmd.Flags().StringVar(&o.OutputFile, "output-file", "", "The generated resources will be copied in the specified file")
+	cmd.Flags().BoolVar(&o.IsHub, "hub", false, "Signal whether the cluster is a hub cluster")
```
qiujian16, line 71: do we need this? could we identify this in the addon controller?

JustinKuli: We didn't have a great way of always identifying the cluster as a hub cluster in the addon-controller. If the cluster's name is `local-cluster`, then it will consider it a hub, but that's the only special case: https://github.com/open-cluster-management-io/governance-policy-addon-controller/blob/main/pkg/addon/policyframework/agent_addon.go#L67

But if I remember right, the registration-operator scripts we use to start a kind cluster can create a self-managed hub that is not named "local-cluster". So we needed a way to specify it in the addon.

Or do you have a suggestion of how we could do it in the addon controller?

qiujian16: we probably could let registration-operator to set a clusterclaim on whether the cluster is hub or not. Or we could add flag  here to add annotations on the addon resource. I'd think self-managed seems to be a special case for certain addons. We may use a more common flag. 

@zhiweiyin318 how do we identify if a cluster is hub in addon-framework today?

dhaiducek: @qiujian16 I agree--if there were a signal that a cluster were a Hub, we'd prefer that.

@zhiweiyin318 Any thoughts here?

zhiweiyin318: > @zhiweiyin318 how do we identify if a cluster is hub in addon-framework today?

we did not identify a hub cluster in addon-framework today. each addon identifies that by themselves, like the name or label of managedCluster.

clusterClaim looks like a better way to do this if it works on hub of hub or multi-hub scenario.

dhaiducek: What's path forward here, then? Are ClusterClaim changes coming soon? If not, I think we'd like to get these changes in so they're captured in the next release (unless there's currently a better way to do this).

qiujian16: I'd rather we remove this flag. self-management is not a very common case I think, is it for the purpose of e2e testing?

I would suggest we do not add this flag, and maybe just doc that user should not add this addon if it is a self-mananged cluster.

dhaiducek: That works for me! I'll remove it and update the doc PR to instruct how to do it manually for self-managed clusters.

mprahl: > I'd rather we remove this flag. self-management is not a very common case I think, is it for the purpose of e2e testing?

I believe that this is a common case for the Policy addon because users want to enforce configuration on the Hub in the same way that they do for managed clusters. That way the Hub isn't a special case for generating compliance reports for meeting various standards and regulatory requirements.
> 
> I would suggest we do not add this flag, and maybe just doc that user should not add this addon if it is a self-mananged cluster.

My vote is to add the flag for now and we can remove it if there is a better alternative later.

qiujian16: this is not convincing to me, since we are adding this flag for ANY addon not only policy. If we think self-managed cluster is not a special case, we should not add special flag for self-managed cluster. It is also error prone, since user can easily enable the addon without the flag in the self-managed cluster. We have to doc it no matter there is a flag or not.

I think we should avoid it to reduce confusion. A better solution is the addon itself detects the property of the cluster, or maybe addon agent can update addon status indicating it is degraded when on a self-managed cluster.

mprahl: @qiujian16 would validation on the flag help so that an error is returned if the flag is used on an addon other than the policy framework?

qiujian16: I think it is an anti-pattern that we have this flag as a special one for a certain addon. We should try to avoid that. Also I am not sure how helpful it is, user can still enable the addon on the self-managed cluster without specifying this flag.

mprahl: @dhaiducek could you have the clusteradm command search for the Policy Propagator deployment on the cluster and if present, treat it as a hub? Then we wouldn't need the flag.

qiujian16: could we just avoid any special code in clusteradm? I really want this part to be as general as possible. We could have another discussion on how this can be done in addon, since I still think the addon should handle the different property of cluster.

qiujian16: I am fine with adding an annotation. we have to add a doc no matter this flag exists or not, having a doc to add an annotation or label on cluster for self-managed cluster feels better to me.

dhaiducek: As discussed here and in Slack, `--hub` has been replaced with `--annotate`.

qiujian16:
/approve
/lgtm
/hold

looks good to me, thanks! hold in case you need others to review.

pkg/cmd/addon/enable/exec.go
```diff
@@ -19,20 +20,32 @@ import (
...
 	NameSpace   string
 	AddonName   string
+	Annotations map[string]string
 }
 
-func NewClusterAddonInfo(cn string, ns string, an string) ClusterAddonInfo {
+func NewClusterAddonInfo(cn string, o *Options, an string) (ClusterAddonInfo, error) {
+	// Parse provided annotations
+	annos := map[string]string{}
+	if len(o.Annotate) > 0 {
```
mprahl, line 33: Optional: You can remove this because you can safely range over a nil slice.

pkg/cmd/addon/enable/cmd.go
```diff
@@ -12,10 +12,24 @@ import (
...
-# Enable addon on to the given managed clusters of the specify namespace
+## Application Manager
+
+# Enable application-manager addon on the given managed clusters in the specified namespace
 %[1]s addon enable --names application-manager --namespace namespace --clusters cluster1,cluster2
-# Enable application-manager addon specified clusters
+# Enable application-manager addon for specified clusters
 %[1]s addon enable --names application-manager --clusters cluster1,cluster2
+
+## Policy Framework
```
mprahl, line 22: Could you please add an example with the annotation for installing on the Hub?

dhaiducek: I was thinking of leaving it off since it's a somewhat special case, but it would be good to have documented in the CLI. I'll add it in.

mprahl (approved):

dhaiducek:
@qiujian16 Would you please re-run the "Go / test" workflow? The failure doesn't appear to be related to my code.

PR #230, opened 2022-05-26
Update: proxy kubectl logs

xuezhaojun:

### DIFF
--- a/pkg/cmd/proxy/kubectl/cmd.go
+++ b/pkg/cmd/proxy/kubectl/cmd.go
@@ -124,10 +124,8 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 			klog.V(4).Infof("kubeconfig file path is %s", tmpKubeconfigFilePath)
 
 			// Using kubectl to access the managed cluster using the above customized kubeconfig
-			result, err := runKubectlCommmand(tmpKubeconfigFilePath, o.kubectlArgs)
-			if err != nil {
-				return errors.Wrap(err, "run kubectl commands failed")
-			}
+			// We are using combinedoutput, so err msg should include in result, no need to handle err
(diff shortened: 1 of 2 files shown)


pkg/cmd/proxy/kubectl/cmd.go
```diff
@@ -124,10 +124,8 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 			klog.V(4).Infof("kubeconfig file path is %s", tmpKubeconfigFilePath)
 
 			// Using kubectl to access the managed cluster using the above customized kubeconfig
-			result, err := runKubectlCommmand(tmpKubeconfigFilePath, o.kubectlArgs)
-			if err != nil {
-				return errors.Wrap(err, "run kubectl commands failed")
-			}
+			// We are using combinedoutput, so err msg should include in result, no need to handle err
+			result, _ := runKubectlCommmand(tmpKubeconfigFilePath, o.kubectlArgs)
```
qiujian16, line 128: why we ignore the err?

xuezhaojun: Because we are using "CombindedOutput",
https://github.com/open-cluster-management-io/clusteradm/blob/f885005d9e3ac1975042156758f7c0aa8ea8e460/pkg/cmd/proxy/kubectl/cmd.go#L256

So all err msg from executing kubectl will be put into the `result` part, the `err` doesn't contain any useful information.

xuezhaojun: For example, the following msg is in `result`:
```
Error from server (Forbidden): pods is forbidden: User "system:serviceaccount:open-cluster-management-managed-serviceaccount:klusterlet" cannot list resource "pods" in API group "" in the namespace "open-cluster-management-agent"
```

PR #232, opened 2022-05-30
update wording to make error clear

jichenjc:
```
# clusteradm delete clusterset default1
Clusterset default1 is already deleted

```

I don't have default1 created.. so it's confusing output


qiujian16 (commented):
/approve
/lgtm

PR #233, opened 2022-05-30
prevent delete of `default` clusterset

### LINKED ISSUE #198: should preventing delete default clusterset
we should preventing delete default clusterset 
as following, allow delete then operator is able to re-create it but seems allow delete default doesn't make sense

```
root@jitest19:~# ./clusteradm get clustersets
NAME      BOUND NAMESPACES   STATUS
default                      1 ManagedClusters selected
root@jitest19:~# ./clusteradm delete clusterset default
Clusterset default is deleted
root@jitest19:~# ./clusteradm get clustersets
root@jitest19:~# ./clusteradm get clustersets
root@jitest19:~# ./clusteradm get clustersets
root@jitest19:~# ./clusteradm get clustersets
NAME      BOUND NAMESPACES   STATUS
default                      1 ManagedClusters selected
root@jitest19:~# ./clusteradm get works

```

jichenjc:
Fix #198

### DIFF
--- a/pkg/cmd/delete/clusterset/exec.go
+++ b/pkg/cmd/delete/clusterset/exec.go
@@ -51,6 +51,12 @@ func (o *Options) runWithClient(clusterClient clusterclientset.Interface,
 	dryRun bool,
 	clusterset string) error {
 
+	// not allow to delete default clusterset
+	if clusterset == "default" {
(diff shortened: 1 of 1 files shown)


qiujian16:
thanks, I think it would also worth to avoid deletion from validating webhook.

FYI @ldpliu 

/approve
/lgtm

PR #245, opened 2022-06-29
Added command to create a sample application

### LINKED ISSUE #108: command to create a sample app
In order to ease the barrier of entry for deploying applications across multicluster using OCM, create a new clusteradm subcommand that generates a sample application that can be deployed into multiple managed clusters. This inspiration comes from the `helm create` command: https://helm.sh/docs/helm/helm_create

I suggest we starts with something easy: `clusteradm create app sample-app`
This will create a folder `sample-app` containing yaml files that upon kubectl apply will perform a helm repo nginx app deployment to all the managed clusters.

What do you think @itdove ?

/assign mikeshng

IlonaShishov:
fixes #108


qiujian16:
Thanks, I think we already have the code to install application addon and enable it. I would think the flow would be

```
clusteradm install hub-addon --names application-manager
clusteradm addon enable --clusters cluster1
clusteradm create sampleapp --namespace xxx
```

To be more general, how about we have a command
```
clusteradm create app --channel helm --chart xxx --namespace xxx --clusters xxx
```

IlonaShishov:
> Thanks, I think we already have the code to install application addon and enable it. I would think the flow would be
> 
> ```
> clusteradm install hub-addon --names application-manager
> clusteradm addon enable --clusters cluster1
> clusteradm create sampleapp --namespace xxx
> ```
> 
> To be more general, how about we have a command
> 
> ```
> clusteradm create app --channel helm --chart xxx --namespace xxx --clusters xxx
> ```

Hi @qiujian16 ,
The only functionality of this new cmd is to deploy a Sample Application to all managed clusters (it does not deploy addons),
the addon install and enable files you are seeing are only for test env setup purposes. without the setup of the application-manager addon on the test cluster it would not be possible to run tests on exec.go.

IlonaShishov:
> could we run an e2e instead of ut? I do not incline having duplicate manifests maintained in the same repo. Or in your test, you could import the existing manifest in the hub addon install package.

Hi @qiujian16,
I have removed the duplicate manifests and instead imported the existing manifests from the hub-addon install package and the addon enable package as per your suggestion, thank you

pkg/cmd/create/sampleapp/exec.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (o *Options) runWithClient(clusterClient clusterclientset.Interface,
+	kubeClient kubernetes.Interface,
+	apiExtensionsClient apiextensionsclient.Interface,
+	dynamicClient dynamic.Interface,
+	dryRun bool) error {
+
+	// Label all managed clusters with clusterset and placement labels
+	err := o.labelManagedClusters(clusterClient)
```
qiujian16, line 77: could we just let user to bind the clusterset at first instead of automatically adding labels and create binding. You could pre-check the binding here and print a warning that user does not have binding in this ns yet.

IlonaShishov: Fixed, removed automatic binding and added pre-check with warning message, Thank you

qiujian16:
/approve
@mikeshng PTAL

Also we may consider a common create app command in the future?

mikeshng (approved):
/lgtm

Thank you for your contribution! This should give us the base we needed to create a new "create app" command or enhance this one to become that.

mikeshng:
@itdove @qiujian16 could you please re-trigger the Go test GitHub action. I took a quick look at the log and it doesn't seem to be related.

mikeshng:
@ycyaoxdu could you please retrigger the CICD? Thanks.

mikeshng:
Thank you! We may need to hold for just a tiny bit more until we tag 0.3.1 due to a critical fix. As soon as 0.3.1 is tagged, this will go in.

PR #248, opened 2022-07-11
Apply

itdove:


qiujian16 (commented):
I am a bit worried about about adding these. Personally I would like the clusteradm to purely focus on ocm related functions (known APIs, known operations). A general template apply func is too much for clusteradm.

itdove:
The apply code get moved to the github.com/stolostron/applier

PR #250, opened 2022-07-13
Add preflight checks

Promacanthus:
When we install a cluster by [`kubeadm`](https://kubernetes.io/docs/reference/setup-tools/kubeadm/implementation-details/#create-the-public-cluster-info-configmap), the deployment tool will create the `cluster-info` ConfigMap  in the `kube-public` namespace. 

However, in some cases, there is no `cluster-info` ConfigMap  in the `kube-public` namespace. 

Under this scenario, the loss of ConfigMap will fail to register managed cluster to hub, and got the following error message.

```shell
can not find the cluster in the cluster-info
```

To resolve this problem, we can do some preflight checks, and print some useful information. 

This commit adds two preflight checks:

1. hub-api-server checker: check whether the server in kubeconfig is a domain name or not. If so, user may add HostAlias into klusterlet.
2. cluster-info checker: check whether the`cluster-info` ConfigMap exists in the `kube-public` namespace or not.

Signed-off-by: Bowen Zhu <Promacanthus@gmail.com>


Promacanthus:
@yue9944882 @ycyaoxdu Please review it when you have time.

pkg/cmd/preflight/testdata/kubeconfig
```diff
@@ -0,0 +1,15 @@
...
+  name: cluster-hub
+contexts:
+- context:
+    cluster: cluster-hub
+    namespace: default
+    user: user-admin
+  name: ocm-hub
+current-context: ocm-hub
+users:
+- name: user-admin
```
ycyaoxdu, line 15: No new line at end of file

pkg/cmd/preflight/testdata/kubeconfig_ip
```diff
@@ -0,0 +1,15 @@
...
+  name: cluster-hub
+contexts:
+- context:
+    cluster: cluster-hub
+    namespace: default
+    user: user-admin
+  name: ocm-hub
+current-context: ocm-hub
+users:
+- name: user-admin
```
ycyaoxdu, line 15: No new line at end of file

pkg/genericclioptions/clusteradm_flags.go
```diff
@@ -23,4 +24,5 @@ func NewClusteradmFlags(f cmdutil.Factory) *ClusteradmFlags {
 func (f *ClusteradmFlags) AddFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&f.DryRun, "dry-run", false, "If set the generated resources will be displayed but not applied")
 	flags.IntVar(&f.Timeout, "timeout", 300, "extend timeout from 300 secounds ")
+	flags.StringVar(&f.Context, "context", "", "The name of the kubeconfig context to use")
```
ycyaoxdu, line 27: `kubectl` have already provide a flag `--context` to specify which context to use, maybe you can't use the flag `context` here? wondering whether this will override the kubectl flag? or do we need to do something to handle this?

ycyaoxdu (changes requested):
`preflight` is not a command to use, so we can not just put it into `pkg/cmd`, since it is used in command `init`, shall we make it a subdirectory of `pkg/cmd/init`?

Promacanthus:
> My Apology for the delay, and thanks a lot for contributing! @Promacanthus
> 
> `preflight` is not a command to use, so we can not just put it into `pkg/cmd`, since it is used in command `init`, shall we make it a subdirectory of `pkg/cmd/init`?

Agreed with you, I will move it into `pkg/cmd/init`.

pkg/genericclioptions/clusteradm_flags.go
Promacanthus: > `kubectl` have already provide a flag `--context` to specify which context to use, maybe you can't use the flag `context` here? wondering whether this will override the kubectl flag? or do we need to do something to handle this?

I did not think about that before. I think we can get context from `kubectl` flag `--context`.

ycyaoxdu (approved):
/lgtm

ycyaoxdu:
looks good to me, any suggestions? @qiujian16

Issue #47, opened 2021-07-14
Race condition where applying the CR happens before the CRD is successfully deployed


sahare:
Getting error when trying to deploy a cluster manager on kind cluster using clusteradm built from the source by following
https://github.com/open-cluster-management-io/clusteradm/blob/main/README.md#from-source

Error:
clusteradm init
E0714 16:01:52.864939  33569 memcache.go:196] couldn't get resource list for operator.open-cluster-management.io/v1: the server could not find the requested resource
Error: no matches for kind "ClusterManager" in version "operator.open-cluster-management.io/v1"

Steps to reproduce:
git clone https://github.com/open-cluster-management-io/clusteradm.git
cd clusteradm
make build
kubectl config use-context kind-hub 
clusteradm init

Investigation by @mikeshng:

I think what is happening is there is for slower environment.
There is a race condition, where its applying the CR before the CRD is successfully deployed.
So:
https://github.com/open-cluster-management-io/clusteradm/blob/main/pkg/cmd/init/exec.go#L127 <- CR apply
gets run too fast before
https://github.com/open-cluster-management-io/clusteradm/blob/main/pkg/cmd/init/exec.go#L92 <- CRD apply
finishes.

sahare:
FYI @mikeshng

PR #276, opened 2022-10-11
bug fix: cr apply before crd registered

ycyaoxdu:


ycyaoxdu:
Related issue: https://github.com/open-cluster-management-io/clusteradm/issues/47
we should apply cr after crd registered successfully.

PR #279, opened 2022-10-14
sync governance-policy-addon-controller deployment image

haoqing0110:
Fix below error when I try to deploy the synchronization components to the managed cluster(s) following steps in
https://open-cluster-management.io/getting-started/integration/policy-framework/#deploy-the-synchronization-components-to-the-managed-clusters.

On managed cluster reports below error:
```
$ oc logs -n open-cluster-management-agent klusterlet-work-agent-858b76c7d-sjhpk
....
E1014 08:17:11.448118       1 base_controller.go:270] "ManifestWorkAgent" controller failed to sync "addon-governance-policy-framework-deploy", err: Deployment.apps "governance-policy-framework" is invalid: spec.template.spec.containers[0].image: Required value
```

Update the image refer to PR https://github.com/open-cluster-management-io/governance-policy-addon-controller/pull/25.

Signed-off-by: haoqing0110 <qhao@redhat.com>

PR #280, opened 2022-10-18
Pin the OCM version for add-ons

dhaiducek:
This PR made us realize that always pulling the `latest` image could cause problems:
- #279 

This pins the version to the latest OCM, but this is configurable if a user wants to pass `--version=latest`.

### DIFF
--- a/pkg/cmd/install/hubaddon/cmd.go
+++ b/pkg/cmd/install/hubaddon/cmd.go
@@ -48,8 +48,9 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	}
 
 	cmd.Flags().StringVar(&o.names, "names", "", "Names of the built-in add-on to install (comma separated). The built-in add-ons are: application-manager, governance-policy-framework")
-	cmd.Flags().StringVar(&o.values.Namespace, "namespace", "open-cluster-management", "Namespace of the built-in add-on to install. Default to open-cluster-management")
+	cmd.Flags().StringVar(&o.values.Namespace, "namespace", "open-cluster-management", "Namespace of the built-in add-on to install. Defaults to open-cluster-management")
 	cmd.Flags().StringVar(&o.outputFile, "output-file", "", "The generated resources will be copied in the specified file")
+	cmd.Flags().StringVar(&o.bundleVersion, "bundle-version", "default", "The image version tag to use when deploying the hub add-on")
 
 	return cmd
 }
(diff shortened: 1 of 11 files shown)


dhaiducek:
/cc @mikeshng

I've included App Lifecycle in this update. Let me know if App would prefer to be left out of this PR and stick with `latest`.

pkg/cmd/install/hubaddon/cmd.go
```diff
@@ -48,8 +48,9 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	}
 
 	cmd.Flags().StringVar(&o.names, "names", "", "Names of the built-in add-on to install (comma separated). The built-in add-ons are: application-manager, governance-policy-framework")
-	cmd.Flags().StringVar(&o.values.Namespace, "namespace", "open-cluster-management", "Namespace of the built-in add-on to install. Default to open-cluster-management")
+	cmd.Flags().StringVar(&o.values.Namespace, "namespace", "open-cluster-management", "Namespace of the built-in add-on to install. Defaults to open-cluster-management")
 	cmd.Flags().StringVar(&o.outputFile, "output-file", "", "The generated resources will be copied in the specified file")
+	cmd.Flags().StringVar(&o.values.OcmVersion, "version", "v0.8.0", "The image version tag to use when deploying the hub add-on. Defaults to the latest released open-cluster-management version.")
```
mprahl, line 53: The text "Defaults to the latest released open-cluster-management version" indicates that it's the latest at the time of `clusteradm` execution but it's really at release time of `clusteradm`.

dhaiducek: > I think you could say it defaults to `v0.8.0` for now.

That's definitely the obvious and simplest answer, haha!

mprahl (approved):
Nice!

dhaiducek:
I'm going to rework this. I was wondering how the `init` function worked, and it already has logic with OCM versions that we should be able to leverage with patterns that we should also be following in our add-on command:
https://github.com/open-cluster-management-io/clusteradm/blob/257c6fd53c6c8573f2e6dafee4c826e192dcbf22/pkg/cmd/init/cmd.go#L53-L54
- Versioning helper: https://github.com/open-cluster-management-io/clusteradm/blob/main/pkg/helpers/version/version.go

dhaiducek:
The `e2e-test` appears to be a flake and should pass if a maintainer would please re-run it.

dhaiducek:
To be clear, we would much prefer that it be included in `v0.4.0` rather than a z-stream.

mikeshng:
/lgtm

I don't have the permission to approve

PR #281, opened 2022-10-20
update changelog

ycyaoxdu:


qiujian16:
should we change version.txt also?

qiujian16:
I would call it 0.4.0

Issue #267, opened 2022-08-30
Enhance the clusteradm to support installing klusterlet in the hosed mode


elgnay:
When joining as cluster to a hub, user is able to set the install mode of the klusterlet to `Hosted` with command similar to
`clusteradm join  …. --mode hosted --managed-cluster-kubeconfig kubeconfig`

PR #289, opened 2022-11-21
feature: add hosted mode option for join command

ycyaoxdu:

### DIFF
--- a/pkg/cmd/join/cmd.go
+++ b/pkg/cmd/join/cmd.go
@@ -13,6 +13,8 @@ import (
 var example = `
 # Join a cluster to the hub
 %[1]s join --hub-token <tokenID.tokenSecret> --hub-apiserver <hub_apiserver_url> --cluster-name <cluster_name>
+# Join a cluster to the hub with hosted mode
+%[1]s join --hub-token <tokenID.tokenSecret> --hub-apiserver <hub_apiserver_url> --cluster-name <cluster_name> --mode hosted --managed-cluster-kubeconfig <path_to_kubeconfig>
 `
 
 // NewCmd ...
@@ -55,5 +57,7 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 		"If true, the installed klusterlet agent will be starting the cluster registration process by "+
 			"looking for the internal endpoint from the public cluster-info in the hub cluster instead of from --hub-apiserver.")
 	cmd.Flags().BoolVar(&o.wait, "wait", false, "If true, running the cluster registration in foreground.")
+	cmd.Flags().StringVarP(&o.mode, "mode", "m", "default", "mode to deploy klusterlet, can be default or hosted")
+	cmd.Flags().StringVar(&o.managedKubeconfigFile, "managed-cluster-kubeconfig", "", "To specify the directory to external managed cluster kubeconfig in hosted mode")
 	return cmd
 }
--- a/pkg/cmd/join/exec.go
+++ b/pkg/cmd/join/exec.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"sync/atomic"
 	"time"
 
@@ -30,6 +31,21 @@ import (
 	"open-cluster-management.io/clusteradm/pkg/helpers/wait"
 )
 
+const (
+	OperatorNamesapce         = "open-cluster-management"
+	DefaultOperatorName       = "klusterlet"
+	InstallModeDefault        = "Default"
+	InstallModeHosted         = "Hosted"
+	KlusterletNamespacePrefix = "open-cluster-management-"
+)
+
+func format(s string) string {
+	if s == "" {
+		return ""
+	}
+	return strings.ToUpper(s[:1]) + strings.ToLower(s[1:])
+}
+
 func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
 	if o.token == "" {
 		return fmt.Errorf("token is missing")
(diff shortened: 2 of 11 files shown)


ycyaoxdu:
related to #267

ycyaoxdu:
use `--mode` to set deploy mode and `--external-managedcluster-kubeconfig` to set external managed cluster kubeconfig.

pkg/cmd/join/exec.go
```diff
@@ -92,6 +97,16 @@ func (o *Options) validate() error {
 		return fmt.Errorf("registry should not be empty")
 	}
 
+	if o.mode != "default" && o.mode != "hosted" {
```
qiujian16, line 100: should set a default value, we should not let user to always set mode

pkg/cmd/join/exec.go
```diff
@@ -114,6 +129,15 @@ func (o *Options) run() error {
 		return err
 	}
 
+	// get ManagedKubeconfig from given file
+	if o.mode == "hosted" {
+		managedConfig, err := os.ReadFile(o.managedKubeconfigFile)
```
qiujian16, line 134: you may also need to check whether it is a valid kubeconfig

pkg/cmd/join/scenario/join/default/klusterlets.cr.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright Contributors to the Open Cluster Management project
+apiVersion: operator.open-cluster-management.io/v1
+kind: Klusterlet
+metadata:
+  name: klusterlet
```
qiujian16, line 5: I do not think the name here should be fixed. multiple hosted klusterlet can run in the same clusters.

pkg/cmd/join/scenario/join/hosted/namespace.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright Contributors to the Open Cluster Management project
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    workload.openshift.io/allowed: "management"
```
qiujian16, line 6: not necessary for this annotation

pkg/cmd/join/scenario/join/hosted/namespace.yaml
```diff
@@ -0,0 +1,7 @@
...
+  name: klusterlet
```
qiujian16, line 7: the namespace should not be fixed

pkg/cmd/join/exec.go
ycyaoxdu: yes, the default value is `default` https://github.com/open-cluster-management-io/clusteradm/pull/289/files#diff-af2f69ba48e16b856a2cce7544ec645881c15d4840f8922fcfb0e49505fb1e39R58

pkg/cmd/join/cmd.go
```diff
@@ -55,5 +55,9 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 		"If true, the installed klusterlet agent will be starting the cluster registration process by "+
 			"looking for the internal endpoint from the public cluster-info in the hub cluster instead of from --hub-apiserver.")
 	cmd.Flags().BoolVar(&o.wait, "wait", false, "If true, running the cluster registration in foreground.")
+	cmd.Flags().StringVarP(&o.mode, "mode", "m", "default", "mode to deploy klusterlet, can be default or hosted")
+	cmd.Flags().StringVar(&o.managedKubeconfigFile, "managed-cluster-kubeconfig", "", "To specify the directory to external managed cluster kubeconfig in hosted mode")
+	cmd.Flags().StringVar(&o.klusterletName, "klusterlet-name", "klusterlet", "To specify klusterlet name(the klusterlet will deploy in the namespace with same name of klusterlet name) in hosted mode")
```
qiujian16, line 60: maybe we can just use cluster name as default value for klusterletname and klusterletnamespace. @zhujian7 would you like to take a look

ycyaoxdu:
#298

pkg/cmd/join/cmd.go
zhujian7: if we use the cluster name and the hosting cluster is the hub cluster, not sure if there will be any issue when detaching the cluster.

zhujian7:
I think we need to take a look at the others related commands.
for example the output of the `clusteradm get token`  contains `please log on spoke and run:`, should we addon more description when it is hosted mode, the join command should be run on the hosting cluster instead of the spoke cluster?

pkg/cmd/join/cmd.go
```diff
@@ -55,5 +55,9 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
...
+	cmd.Flags().StringVar(&o.klusterletNamespace, "klusterlet-namespace", "klusterlet", "To specify klusterlet namespace on managed cluster in hosted mode")
```
zhujian7, line 61: could you give more explanation of the klusterlet-namespace?

ycyaoxdu:
Talked with @zhujian7 , the result is: 
1. we can not directly set the flag `--klusterlet-name` to `<managedcluster name>` . Here is a case offered by @zhujian7 : if we use `clusterA` as `hub cluster` and `management cluster` and set the flag `--klusterlet-name` to `<managedcluster name>`, the klusterlet name(as well as the agent namespace) is  same as managedcluster name, this will leading to the two group of resources(managedcluster related and klusterlet related) stores in the same namespace, and while detaching, the namespace will be deleted, causing the managedcluster resources will be deleted unexpectedly;
2. the flag `--klusterlet-namespace` can be removed, the name of this flag is misleading and we can just set the value to `open-cluster-management-<klusterlet name>` in our code.
----
@qiujian16 do you have any suggestion?

qiujian16:
maybe the combination of the cluster name as a prefix and a spec hash of klusterlet?

pkg/cmd/join/exec.go
```diff
@@ -49,15 +54,27 @@ func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
 	// convert mode string to lower
 	o.mode = strings.ToLower(o.mode)
 
+	// In default deploy mode, klusterlet is deployed in operator namespace "open-cluster-management", the agents(such as registration-agent) are
+	// deployed in agent namespace "open-cluster-management-agent".
+	// In hosted deploy mode, klusterlet and agents are both deployed in agent namespace(the agent namespace is on the management cluster, and share
+	// name with klusterlet); the klusterletNamespace is still on the managed cluster, the name is "open-cluster-management-<klusterlet-name>".
+	klusterletName := "open-cluster-management"
```
zhujian7, line 61: this is a little confused, the default klusterlet name is "open-cluster-management"?

pkg/cmd/join/exec.go
```diff
@@ -45,12 +61,41 @@ func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
...
+	// In default mode, agentNamespace is "open-cluster-management-agent", klusterletNamespace refers to agentNamespace, all of these three namesapces are on the managed cluster;
+	// In hosted mode, operatorNamespace is on the management cluster, agentNamesapce is "<cluster name>-<6-bit random string>" on the management cluster, and the klusterletNamespace is "open-cluster-management-<agentNamespace>" on the managed cluster.
+
+	// values for default mode
+	klusterletName := DefaultOperatorName
+	agentNamespace := KlusterletNamespacePrefix + "agent"
+	klusterletNamespace := agentNamespace
+	if o.mode == InstallModeHosted {
+		// add hash suffix to avoid conflict
+		klusterletName = o.clusterName + "-" + helpers.RandStringRunes_az09(6)
```
qiujian16, line 81: if we use random number, let's just do klusterletName = klusterletName + "-hosted-" + helpers.RandStringRunes_az09(6)

WDYT?

ycyaoxdu: @qiujian16 The reason for the klusterlet name is user can easily corresponds klusterlet CR to managed cluster at first glance, if there are many klusterlets on mangement cluster. 
But if isn't important, I think `klusterletName + "-hosted-" + helpers.RandStringRunes_az09(6)` is absolutely OK. 
---

BTW, what should klusterletName be? is it fixed to `klusterlet` or just equals to `o.clusterName`?

qiujian16: I think we have cluster name in spec of klusterlet? The klusterlet name could be a static `klusterlet`

qiujian16 (commented):
/approve

It worth a demo on community meeting

ycyaoxdu:
> 

@qiujian16 maybe after we've done more jobs in https://github.com/open-cluster-management-io/clusteradm/issues/298 ?

pkg/cmd/join/preflight/checks.go
```diff
@@ -49,6 +55,40 @@ func (c HubKubeconfigCheck) Name() string {
...
+	ManagedKubeconfigFile string
+}
+
+func (c DeployModeCheck) Check() (warningList []string, errorList []error) {
+	if c.Mode != InstallModeDefault && c.Mode != InstallModeHosted {
+		return nil, []error{errors.New("deploy mode should be default or hosted")}
+	}
+	if c.Mode == InstallModeDefault {
+		if c.ManagedKubeconfigFile != "" {
+			return nil, []error{errors.New("--managed-cluster-kubeconfig should not be set in default deploy mode")}
```
zhujian7, line 70: I am not sure if we do need this check.

ycyaoxdu: I think we should not set the --managed-cluster-kubeconfig with default mode. Can it be set in default mode? @zhujian7

zhujian7: I was thinking just ignoring it in default mode, but reporting an error also makes sense.

zhujian7:
LGTM

Issue #274, opened 2022-10-06
Klusterlet CR generated from clusteradm join command contains invalid externalServerURLs for Oracle Kubernetes Engine cluster


mikeshng:
In an Oracle Kubernetes Engine cluster, the cluster-info config map:
```
kubectl -n kube-public get cm cluster-info -o yaml
```
contains
```
server: 1.2.3.4:6443 # without the https://
```
this might have cause the clusteradm to generate a Klusterlet CR with the spec of 
```
    externalServerURLs:
    - url: 1.2.3.4:6443 # without the https://
```

This fails the spoke agent validation in https://github.com/open-cluster-management-io/registration/blob/release-0.8/pkg/spoke/spokeagent.go#L435-L442
causing the managed cluster not to proceed with the registration process.

The workaround is modify the klusterlet CR by appending the `https://` and the registration went through successfully.

```
kubectl version: v1.23.7+1
clusteradm version:
client          version :v0.3.1
server release  version :v1.23.4
default bundle  version :0.8.0
Kubernetes version: v1.23.4
Distribution: Oracle Kubernetes Engine
```

Reported by @hyder

hyder:
Hi,

Sorry for the delay.

Please find troubleshooting steps for this issue:

```
[opc@mc-operator ~]$ kubectl -n open-cluster-management-agent get pods
NAME                                           READY   STATUS             RESTARTS      AGE
klusterlet-registration-agent-dd8d4889-79ngn   0/1     CrashLoopBackOff   2 (15s ago)   49s

kubectl -n open-cluster-management-agent logs -f klusterlet-registration-agent-dd8d4889-79ngn

F1014 10:47:36.421943       1 spokeagent.go:132] "10.23.0.22:6443" is invalid
goroutine 1 [running]:
k8s.io/klog/v2.stacks(0x1)
        k8s.io/klog/v2@v2.60.1/klog.go:860 +0x8a
k8s.io/klog/v2.(*loggingT).output(0x3b298c0, 0x3, 0x0, 0xc0007a78f0, 0x1, {0x2eab442?, 0x7f8b8019cf18?}, 0x10?, 0x0)
        k8s.io/klog/v2@v2.60.1/klog.go:825 +0x686
k8s.io/klog/v2.(*loggingT).printDepth(0x3b298c0, 0x282ecd0?, 0x0, {0x0, 0x0}, 0x2828ae0?, {0xc000a3cbb0?, 0x1, 0x1})
[...]

PR #291, opened 2022-11-29
refactor join command and add preflight check

ycyaoxdu:

### DIFF
--- a/pkg/cmd/join/exec.go
+++ b/pkg/cmd/join/exec.go
@@ -4,6 +4,7 @@ package join
 import (
 	"context"
 	"fmt"
+	"os"
 	"sync/atomic"
 	"time"
 
@@ -21,8 +22,10 @@ import (
 	clientcmdapiv1 "k8s.io/client-go/tools/clientcmd/api/v1"
 	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/cmd/util"
+	"open-cluster-management.io/clusteradm/pkg/cmd/join/preflight"
 	"open-cluster-management.io/clusteradm/pkg/cmd/join/scenario"
 	"open-cluster-management.io/clusteradm/pkg/helpers"
+	preflightinterface "open-cluster-management.io/clusteradm/pkg/helpers/preflight"
 	"open-cluster-management.io/clusteradm/pkg/helpers/printer"
 	"open-cluster-management.io/clusteradm/pkg/helpers/version"
 	"open-cluster-management.io/clusteradm/pkg/helpers/wait"
@@ -90,6 +93,16 @@ func (o *Options) validate() error {
 	}
 	if len(o.registry) == 0 {
 		return fmt.Errorf("registry should not be empty")
+	}
+
+	// preflight check
+	if err := preflightinterface.RunChecks(
+		[]preflightinterface.Checker{
+			preflight.KlusterletApiserverCheck{
+				KlusterletApiserver: o.values.Klusterlet.APIServer,
+			},
+		}, os.Stderr); err != nil {
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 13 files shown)


ycyaoxdu:
related issue: https://github.com/open-cluster-management-io/clusteradm/issues/274

pkg/cmd/join/exec.go
```diff
@@ -92,6 +95,16 @@ func (o *Options) validate() error {
 		return fmt.Errorf("registry should not be empty")
 	}
 
+	// preflight check
+	if err := preflightinterface.RunChecks(
+		[]preflightinterface.Checker{
+			preflight.ApiHostIPCheck{
+				ApiHostIP: o.values.Hub.APIServer,
```
qiujian16, line 102: is this the value read from kube-public config map?

ycyaoxdu: done

pkg/cmd/join/exec.go
```diff
@@ -92,6 +95,16 @@ func (o *Options) validate() error {
 		return fmt.Errorf("registry should not be empty")
 	}
 
+	// preflight check
+	if err := preflightinterface.RunChecks(
+		[]preflightinterface.Checker{
+			preflight.KlusterletApiserverCheck{
```
qiujian16, line 101: are you sure it works, we read and set this in complete. I think validate is run before complete?

ycyaoxdu: @qiujian16 seems complete runs first, then validate runs: https://github.com/open-cluster-management-io/clusteradm/blob/main/pkg/cmd/join/cmd.go#L33

qiujian16: hrm, with that I think we need to do preflight in complete. The message here is not correct, because apiserver can also be inputed by user I think?

pkg/cmd/join/preflight/checks.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, []error{errors.New("error cluster length")}
+	}
+
+	// validate apiserver foramt
+	if !validAPIHost(c.Config.Clusters[0].Cluster.Server) {
+		return nil, []error{errors.New("--hub-apiserver should start with http:// or https://")}
+	}
+	// validate ca
+	if c.Config.Clusters[0].Cluster.CertificateAuthorityData == nil {
+		return []error{errors.New("no ca detected, creating hub kubeconfig without ca")}, nil
```
qiujian16, line 46: is this an error or warning?

pkg/cmd/join/exec.go
```diff
@@ -79,41 +121,30 @@ func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
...
-	}
-	if len(o.registry) == 0 {
-		return fmt.Errorf("registry should not be empty")
+	// preflight check
+	if err := preflightinterface.RunChecks(
+		[]preflightinterface.Checker{
+			preflight.HubKubeconfigCheck{
+				Config: o.HubConfig,
+			},
+			preflight.KlusterletApiserverCheck{
```
qiujian16, line 130: nit: should we put "check if it is hub" and "check if it is spoke" also as preflight check. Such that this can be used in each cmd? Won't be a blocker for this pr, but worth thinking.

pkg/cmd/join/preflight/checks.go
ycyaoxdu: this is an warning telling the user that the client has built up with no ca data. I think the user should make the decision whether use --ca-file to specify a ca file himself/herself in this condition @qiujian16

qiujian16: but we return an error here.

pkg/cmd/join/preflight/checks.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/pkg/errors"
+	clientcmdapiv1 "k8s.io/client-go/tools/clientcmd/api/v1"
+	"open-cluster-management.io/clusteradm/pkg/helpers"
+)
+
+type KlusterletApiserverCheck struct {
+	KlusterletApiserver string
+}
+
+func (c KlusterletApiserverCheck) Check() (warningList []error, errorList []error) {
```
qiujian16, line 16: I see what's going on here. But you may want warnings as a []string to avoid confusion.

ycyaoxdu: done

PR #296, opened 2022-12-23
Fix: preflight check should return no error when configmap cluster-info not exist.

xuezhaojun:


ycyaoxdu (commented):
My apology for the delay
/lgtm

ycyaoxdu (commented):
My apology for the delay
/lgtm

PR #297, opened 2023-01-05
Fix: warning when klusterletApiServer doesn't start with `https://`

xuezhaojun:

Fixes: https://github.com/open-cluster-management-io/clusteradm/issues/274

### DIFF
--- a/pkg/cmd/join/exec.go
+++ b/pkg/cmd/join/exec.go
@@ -109,6 +109,9 @@ func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
 	if err != nil {
 		klog.Warningf("Failed looking for cluster endpoint for the registering klusterlet: %v", err)
 		klusterletApiserver = ""
+	} else if !preflight.ValidAPIHost(klusterletApiserver) {
+		klog.Warningf("ConfigMap/cluster-info.data.kubeconfig.clusters[0].cluster.server field [%s] in namespace kube-public should start with http:// or https://", klusterletApiserver)
+		klusterletApiserver = ""
 	}
 	o.values.Klusterlet.APIServer = klusterletApiserver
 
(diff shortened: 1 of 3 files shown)


pkg/cmd/join/exec.go
```diff
@@ -109,6 +109,9 @@ func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
 	if err != nil {
 		klog.Warningf("Failed looking for cluster endpoint for the registering klusterlet: %v", err)
 		klusterletApiserver = ""
+	} else if !preflight.ValidAPIHost(klusterletApiserver) {
```
ycyaoxdu, line 112: We already have a check for this, why do we check it here?

xuezhaojun: Yes, there are 2 pieces of code I think are relevant to this question:
1. https://github.com/open-cluster-management-io/registration/blob/release-0.8/pkg/spoke/spokeagent.go#L435-L442： we can see `SpokeExternalServerURLs` could be an empty list.
2. https://github.com/open-cluster-management-io/clusteradm/blob/55a00be91b99fae2641c17634a126382b84af0d2/pkg/cmd/join/scenario/join/klusterlets.cr.yaml#L12: we can see the `.Klusterlet.APIServer` is an optional arg for `join` process.

So what we want here is: when configmap `cluster-info` of a managed cluster can not provide a valid URL, we just give the user a warning msg and skip this arg, instead of returning an error and exiting.

ycyaoxdu: @xuezhaojun Yes, it's definitely right, and we should remove the corresponding preflight check because it will be useless. Could you please also do that in this PR?

xuezhaojun: Sure, preflight removed.

ycyaoxdu (commented):
/lgtm

PR #301, opened 2023-01-20
Add addon OWNERS files

dhaiducek:
Signed-off-by: Dale Haiducek <19750917+dhaiducek@users.noreply.github.com>

### DIFF
--- a/pkg/cmd/addon/OWNERS
+++ b/pkg/cmd/addon/OWNERS
@@ -1,0 +1,12 @@
+approvers:
+- dhaiducek
+- gparvin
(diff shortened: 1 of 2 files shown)


mikeshng:
This might be too many approvers to be onboarded all at once. I am ok if my name is removed and maybe we should only have at most 2 from policy sig.  

FYI @qiujian16

dhaiducek:
@mikeshng Would you please re-lgtm this? It looks like the first review didn't take.

mikeshng (approved):
/lgtm

PR #331, opened 2023-04-13
Add cmd to create placement

qiujian16:


pkg/cmd/create/placement/cmd.go
```diff
@@ -0,0 +1,66 @@
...
+%[1]s create placement test --label-selectors region=apac
+
+# Create a placement with count.
+%[1]s create placement test --count 2
+
+# Create a placement with clustersets.
+%[1]s create placement test --clustersets set1
+
+# Create a placement with clustersets prioritizers
+%[1]s create placement test --prioritizers BuiltIn:Steady:3,BuiltIn:ResourceAllocatableCPU:2
```
haoqing0110, line 25: Does here need an example for using addon score?

pkg/cmd/create/placement/exec.go
```diff
@@ -0,0 +1,165 @@
...
+		return err
+	}
+
+	if !o.Overwrite {
+		return fmt.Errorf("placement %s already exists", placement.Name)
+	}
+
+	placementOrigin.Spec = placement.Spec
+	_, err = clusterClient.ClusterV1beta1().Placements(o.Namespace).Update(context.TODO(), placementOrigin, metav1.UpdateOptions{})
+
```
haoqing0110, line 163: How about providing a hit on how to check placement and placmentdecision after applying the resource successfully.

PR #332, opened 2023-04-17
Rm applier

qiujian16:


ycyaoxdu:
Looks like this PR did these things in general:
1.add reader, use reader to apply resources; remove strolostron/applier
2.refactor addon enable command
3.add AddonManager in Init command
4.add DefaultSpokeWorkFeatureGates
5.add genericclioptions.IOStreams to command option
6.update sets.set to use generic 

all looks good to me

PR #333, opened 2023-04-18
Remove applier deps from tests

qiujian16:
This is futher to remove deps on previous applier code


pkg/cmd/upgrade/clustermanager/exec.go
ycyaoxdu, line 1: maybe `err = r.Apply(init_scenario.Files, o.values, "join/operator.yaml")` at line 96 also need to be fixed?

Issue #323, opened 2023-03-16
[Umbrella Issue] support start multicluster-controlplane 


ycyaoxdu:
- [x] enhance init command to support start a multicluster-controlplane
- [x] enhance join command to support join a multicluster-controlplane agent

related comments: 
https://github.com/open-cluster-management-io/OCM/issues/126#issuecomment-1449216276
https://github.com/open-cluster-management-io/OCM/issues/126#issuecomment-1457779163

ycyaoxdu:
Are there any suggestions on how to define the command architect? 
---

@clyang82 has proposed that we place multicluster-controlplane related commands into one sub-command:
```
clusteradm controlplane init/join xxx
```

---
And I think we can also support multicluster-controlplane as an option in `clusteradm init` and `clusteradm join`, like:
```
clusteradm init/join --controlplane xxx
```

---

WDYT? @qiujian16

PR #336, opened 2023-05-11
allow init hub by deploy multicluster controlplane

ycyaoxdu:


ycyaoxdu:
https://github.com/open-cluster-management-io/clusteradm/issues/323

qiujian16:
@ycyaoxdu since we already be able to install the controlplane using helm chart, I am thinking maybe we just need to ensure we can use clusteradm to join. WDYT?

clyang82:
for the users who are using clusteradm to build their cluster, maybe keep the same experience is better. That means the users can use clusteradm init a controlplane and then use clusteradm join to this controlplane.

clyang82:
I do not expect to maintain the duplicated manifests in different places. The manifests should be maintained in multicluster-controlplane repo. the clusteradm command depends on the chart to init the controlplane. Is it doable or does it make sense?

qiujian16 (commented):
we need to have a better code structure if we intend to use helm to install from clusteradm.

e.g.
the flags for singleton and values should be put int the separated package. For values, I think it needs to be extensible. maybe it should be "--values x=x1,y=y1"

I think it is better if we can construct a helm install interface, so we can reuse it in the future for other components like addons.

qiujian16:
/approve
/lgtm
/hold

I seems ok in general, but I cannot go through it in detail since it is really large.
will merge it at first and resolve the following issue

clyang82:
I am fine with the changes. Let us merge it as an initial version. we can have the following PRs to address the issues.

PR #338, opened 2023-05-16
allow install regiatration agent using multicluster controlplane in join command

ycyaoxdu:


ycyaoxdu:
https://github.com/open-cluster-management-io/clusteradm/issues/323

pkg/cmd/join/exec.go
```diff
@@ -213,84 +255,129 @@ func (o *Options) validate() error {
...
+			)
+		}
+		files = append(files,
+			"join/namespace_agent.yaml",
+			"join/bootstrap_hub_kubeconfig.yaml",
+		)
 
-	klusterletNamespace := o.values.Klusterlet.KlusterletNamespace
-	agentNamespace := o.values.Klusterlet.AgentNamespace
+		if o.mode == InstallModeHosted {
```
qiujian16, line 329: hosted mode should be independent with using standalone agent or not, I think it should be possible that use hosted standalone agent, right?

pkg/cmd/join/cmd.go
```diff
@@ -48,19 +49,39 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
...
+	klusterletSet.StringVar(&o.registry, "image-registry", "quay.io/open-cluster-management", "The name of the image registry serving OCM images.")
+	klusterletSet.StringVar(&o.bundleVersion, "bundle-version", "default",
+		"The version of predefined compatible image versions (e.g. v0.6.0). Defaults to the latest released version. You can also set \"latest\" to install the latest development version.")
+	klusterletSet.StringVar(&o.managedKubeconfigFile, "managed-cluster-kubeconfig", "", "To specify the directory to external managed cluster kubeconfig in hosted mode")
+	_ = klusterletSet.SetAnnotation("image-registry", "mode-orinted", []string{"default", "hosted"})
+	_ = klusterletSet.SetAnnotation("bundle-version", "mode-orinted", []string{"default", "hosted"})
+	_ = klusterletSet.SetAnnotation("managed-cluster-kubeconfig", "mode-orinted", []string{"hosted"})
+
+	//controlplaneAgentSet contains the flags for deploy controlplane agent
+	controlplaneAgentSet := pflag.NewFlagSet("controlplane-agent", pflag.ExitOnError)
```
qiujian16, line 77: controlplane-agent is confusing.  I would use standalone

pkg/cmd/join/cmd.go
```diff
@@ -48,19 +49,39 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
...
+	joinSet.StringVar(&o.clusterName, "cluster-name", "", "The name of the joining cluster")
+	joinSet.BoolVar(&o.forceHubInClusterEndpointLookup, "force-internal-endpoint-lookup", false,
 		"If true, the installed klusterlet agent will be starting the cluster registration process by "+
 			"looking for the internal endpoint from the public cluster-info in the hub cluster instead of from --hub-apiserver.")
-	cmd.Flags().BoolVar(&o.wait, "wait", false, "If true, running the cluster registration in foreground.")
-	cmd.Flags().StringVarP(&o.mode, "mode", "m", "default", "mode to deploy klusterlet, can be default or hosted")
-	cmd.Flags().StringVar(&o.managedKubeconfigFile, "managed-cluster-kubeconfig", "", "To specify the directory to external managed cluster kubeconfig in hosted mode")
+	joinSet.BoolVar(&o.wait, "wait", false, "If true, running the cluster registration in foreground.")
+	joinSet.StringVar(&o.outputFile, "output-file", "", "The generated resources will be copied in the specified file")
+	joinSet.StringVarP(&o.mode, "mode", "m", "default", "mode to install OCM agent, can be default, hosted or controlplane-agent")
```
qiujian16, line 64: I do not think hosted and controlplane-agent should be in the same flag

pkg/cmd/join/exec.go
```diff
@@ -213,84 +237,129 @@ func (o *Options) validate() error {
...
+				return err
+			}
+		}
+
+		files := []string{}
 
-	if o.mode == InstallModeHosted {
 		files = append(files,
-			"join/hosted/external_managed_kubeconfig.yaml",
+			"standalone/bootstrap_hub_kubeconfig.yaml",
```
qiujian16, line 267: why this is different from the one for join command?

pkg/cmd/join/exec.go
```diff
@@ -213,84 +237,129 @@ func (o *Options) validate() error {
...
-			"join/service_account.yaml",
-			"join/cluster_role.yaml",
-			"join/cluster_role_binding.yaml",
-		)
-	}
-	files = append(files,
-		"join/namespace_agent.yaml",
-		"join/bootstrap_hub_kubeconfig.yaml",
-	)
+	if o.standalone {
```
qiujian16, line 247: I'd we put standalone deploy in a separated functions and return fast. this big if-else section could be confusing

pkg/cmd/join/cmd.go
```diff
@@ -62,5 +62,7 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	cmd.Flags().BoolVar(&o.wait, "wait", false, "If true, running the cluster registration in foreground.")
 	cmd.Flags().StringVarP(&o.mode, "mode", "m", "default", "mode to deploy klusterlet, can be default or hosted")
 	cmd.Flags().StringVar(&o.managedKubeconfigFile, "managed-cluster-kubeconfig", "", "To specify the directory to external managed cluster kubeconfig in hosted mode")
+	cmd.Flags().BoolVar(&o.standalone, "standalone", false, "If true, deploy standalone controlplane agent instead of klusterlet")
```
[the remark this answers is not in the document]
clyang82: we should not call at as `standalone`. it is misleading. maybe we just use `multicluster-controlplane` instead.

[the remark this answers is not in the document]
qiujian16: IMHO, it is hard to understand what is multicluster-controlplane. It is a project name rather than how we deploy the agent. And in the long term, we probably will build the standalone agent directly from the ocm repo.

PR #340, opened 2023-05-21
Add install hub-addon argocd-pull-manager for ArgoCD pull integration

mikeshng:


qiujian16:
I am wondering whether it is better we should put this in the helm chart instead. I am concerned on how to maintain this in the cmd in the long term.

mikeshng:
closing this PR as helm chart is the prefer option.

PR #343, opened 2023-05-25
update changelog for v0.6.0

ycyaoxdu:


pkg/helpers/version/version.go
```diff
@@ -105,6 +105,15 @@ func GetVersionBundle(version string) (VersionBundle, error) {
 		PolicyAddon:  "v0.10.0",
 	}
 
+	versionBundleList["0.11.0"] = VersionBundle{
+		Registration: "v0.11.0",
+		Placement:    "v0.11.0",
+		Work:         "v0.11.0",
+		Operator:     "v0.11.0",
```
qiujian16, line 112: there is also an addon-manager we need to set

ycyaoxdu: done

PR #346, opened 2023-06-12
ci: prevent fossa workflow from runing and failing when syncing forks

TomerFi:
Signed-off-by: Tomer Figenblat <tfigenbl@redhat.com>


ycyaoxdu:
@qiujian16 PTAL

Issue #347, opened 2023-06-12
proxy subcommand doesn't work for the latest cluster-proxy


yana1205:
Hi, I am going through https://open-cluster-management.io/getting-started/integration/cluster-proxy/ and facing the following error at the step of "Command-line tools" in "Usage" section.
```
$ clusteradm proxy health
Error: failed getting managedproxyconfiguration for cluster-proxy: resource name may not be empty
```

Probably this issue is relating with this comment (https://github.com/open-cluster-management-io/clusteradm/blob/main/pkg/cmd/proxy/health/exec.go#L106) so will be fixed sooner or later but let me open it.

Now I am using cluster-proxy v0.3.0 and it seems that cluster-proxy doesn't return any value in `AddOnConfiguration.CRName` for clustermanagementaddons because probably `addOnConfiguration` is deprecated.
```
$ kubectl get clustermanagementaddons
NAME            DISPLAY NAME    CRD NAME
cluster-proxy   cluster-proxy  
```

If I switch to use `supportedConfigs[0].defaultConfig.name` field (the value is actually `cluster-proxy`) at https://github.com/open-cluster-management-io/clusteradm/blob/main/pkg/cmd/proxy/health/exec.go#L109 as the comment indicates, it turns into working. 
```
$ go run ./cmd/clusteradm/clusteradm.go proxy health                                            
CLUSTER NAME    INSTALLED    AVAILABLE    PROBED HEALTH    LATENCY
cluster1        True         True         True             16.01925ms
```

### Environment
clusteradm
```
$ clusteradm version
client          version :v0.6.0
server release  version :v1.26.0
default bundle  version :0.11.0
```

cluster-proxy addon: v0.3.0
managed-serviceaccount addon: v0.3.0
```
$ helm -n open-cluster-management-addon ls
NAME                    NAMESPACE                       REVISION        UPDATED                                 STATUS          CHART                           APP VERSION
cluster-proxy           open-cluster-management-addon   1               2023-06-12 17:42:56.936825 +0900 JST    deployed        cluster-proxy-0.3.0             1.0.0      
managed-serviceaccount  open-cluster-management-addon   1               2023-06-12 18:05:47.798653 +0900 JST    deployed        managed-serviceaccount-0.3.0    1.0.0      
```

yana1205:
Ah, sorry. I was missing this comment.
[...]

PR #348, opened 2023-06-19
Fix: using deprecated field `AddOnConfiguiration` causing `proxy health` not work.

xuezhaojun:
Fixes: https://github.com/open-cluster-management-io/clusteradm/issues/347

### DIFF
--- a/pkg/cmd/proxy/health/exec.go
+++ b/pkg/cmd/proxy/health/exec.go
@@ -34,6 +34,7 @@ import (
 	"open-cluster-management.io/cluster-proxy/pkg/common"
 	"open-cluster-management.io/cluster-proxy/pkg/generated/clientset/versioned"
 	"open-cluster-management.io/cluster-proxy/pkg/util"
+	"open-cluster-management.io/clusteradm/pkg/config"
 	konnectivity "sigs.k8s.io/apiserver-network-proxy/konnectivity-client/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/cmd/proxy/health/exec.go
```diff
@@ -103,10 +103,8 @@ func (o *Options) run(streams genericclioptions.IOStreams) error {
 		return errors.Wrapf(err, "failed initializing proxy api client")
 	}
 
-	// TODO: fix this deprecated field AddOnConfiguration
-	// nolint:staticcheck
 	proxyConfig, err := proxyClient.ProxyV1alpha1().ManagedProxyConfigurations().
-		Get(context.TODO(), clusterAddon.Spec.AddOnConfiguration.CRName, metav1.GetOptions{})
+		Get(context.TODO(), "cluster-proxy", metav1.GetOptions{})
```
xuezhaojun, line 107: The name of managedproxyconfiguration is always `cluster-proxy` if we deploy addon via clusteradm: https://github.com/open-cluster-management-io/cluster-proxy/blob/2a197b382114b674b76a362146a7259aed164424/charts/cluster-proxy/templates/managedproxyconfiguration.yaml#L4

qiujian16, line 107: set it as a const

xuezhaojun:
/assign @qiujian16 

Rebase, PTAL, thanks!

Issue #344, opened 2023-05-29
Misleading preflight error message


aii-nozomu-oki:
## clusteradm version

latest(eafca6db83b6a3aa56323a1de41ad167f48f5cca)

## How to reproduce

```
$ go run cmd/clusteradm/clusteradm.go join --hub-token test --hub-apiserver https://8.8.8.8 --cluster-name test
W0529 02:44:00.078607   86720 exec.go:160] Failed looking for cluster endpoint for the registering klusterlet: configmaps "cluster-info" not found
Preflight check: HubKubeconfig check Failed with 0 warnings and 1 errors
Preflight check: DeployMode Check Passed with 0 warnings and 0 errors
Preflight check: ClusterName Check Passed with 0 warnings and 0 errors
Error: [preflight] Some fatal errors occurred:
	[ERROR HubKubeconfig check]: no ca detected, creating hub kubeconfig without ca

exit status 1
```

As we know, 8.8.8.8 is not Kubernetes API server, and this is due to a mistake in `--hub-apiserver`.
It would be useful to be able to detect that the target is not Kubernetes API server.

The above is an extreme example, but in my case, I had specified the wrong port.

PR #349, opened 2023-06-20
update preflight check

ycyaoxdu:
https://github.com/open-cluster-management-io/clusteradm/issues/344


ycyaoxdu:
if an invalid `--hub-apiserver` is specified , the error would be:
```
Error: Get "https://127.0.0.1:46619/api/v1/namespaces/kube-public/configmaps/cluster-info": dial tcp 127.0.0.1:46619: connect: connection refused
```

PR #351, opened 2023-06-30
Fix: e2e fails because not support latest bundles.

xuezhaojun:
Fixes: 
* https://github.com/open-cluster-management-io/clusteradm/actions/runs/5423857040/jobs/9862392036?pr=348
* https://github.com/open-cluster-management-io/clusteradm/actions/runs/5319342206/jobs/9631726852?pr=349

---

In `clusteradm` e2e, we should depend on a stable version of bundles, otherwise any bugs or new features imported in "registration-operator", "registration", "work" or other repos, will **break or block the development of clusteradm**.

**Recommended practice**: we can make a task to "Make clusteradm support 0.12.0" when we want to release 0.12.0 binaries.

Because **we can't always promise** the compatibility between latest `clusteradm` and the latest bundles, which would exhaust us.  

So any developer who uses `clusteradm` to deploy `latest` bundles should know **the risk** that there might be compatibility issues.

### DIFF
--- a/pkg/cmd/init/scenario/init/clustermanager_cluster_role.yaml
+++ b/pkg/cmd/init/scenario/init/clustermanager_cluster_role.yaml
@@ -1,4 +1,5 @@
 # Copyright Contributors to the Open Cluster Management project
+# Copied from https://github.com/open-cluster-management-io/ocm/blob/main/deploy/cluster-manager/config/rbac/cluster_role.yaml
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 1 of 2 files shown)


qiujian16:
what if we add a new feature in the latest version, and try to enable it in clusteradm?

xuezhaojun:
> what if we add a new feature in the latest version, and try to enable it in clusteradm? I think the root cause is we duplicate the deploy manifests in clusteradm.

That makes sense, I can fix the inconsistent part in this PR, but finally, I think we should have a way to trigger the e2e of clusteradm when manifests of registration, work, operator changes? That would be a motivation to migrate clusteradm into ocm repo.

xuezhaojun:
> what if we add a new feature in the latest version, and try to enable it in clusteradm? I think the root cause is we duplicate the deploy manifests in clusteradm.

Another thought of this is, maybe we can release a mirror version(for example, registration-operator 0.11.1) and update the clusteradm default bundle setting in e2e.

qiujian16:
Yes, I think we need to some how trigger clusteradm test either periodically or per PR in ocm.

xuezhaojun:
> Yes, I think we need to some how trigger clusteradm test either periodically or per PR in ocm. But for this PR, I'd rather we do not change to default, but fix the permission issue

Got it.

Issue #350, opened 2023-06-22
clusteradm get klusterlet-info fails even with managed cluster context


anandf:
I am trying to get the `klusterlet-info` of a managed cluster by trying to run the command `clusteradm get klusterlet-info` but the command fails with an error message `

Version details
```
clusteradm version
client		version	:v0.6.0
server release	version	:v1.26.3
default bundle	version	:0.11.0
```

Steps to reproduce:
- Run the `local-up.sh` script to create 3 KinD clusters (1 hub and 2 managed clusters)
```
curl -L https://raw.githubusercontent.com/open-cluster-management-io/OCM/main/solutions/setup-dev-environment/local-up.sh | bash
```
- Run the following command to get the `klusterlet` details.
```
multicluster clusteradm get klusterlet-info --context kind-cluster2
```
```
# k ctx kind-cluster2
Switched to context "kind-cluster2".
# clusteradm get klusterlet-info
Error: managed cluster oriented command should not running against non-managed cluster
# multicluster clusteradm get klusterlet-info --context kind-cluster2
Error: managed cluster oriented command should not running against non-managed cluster
```
As a workaround, I am able to run the below command and get the required details.
```
oc get klusterlet -o yaml
```

There are no `ClusterClaim` CRs in the managed cluster which is used to determine if a given context maps to  a hub or managed cluster.
```
# k get clusterclaim -A 
No resources found
```

mikeshng:
https://github.com/open-cluster-management-io/clusteradm/blob/v0.6.0/pkg/helpers/check/check.go#L57-L70

Maybe the issue is the code is looking for a v1 version of clusterclaim but clusteclaim is still on v1alpha1? 

CC @ycyaoxdu

anandf:
```
 # kubectl api-resources | grep clusterclaim
clusterclaims                                        cluster.open-cluster-management.io/v1alpha1   false        ClusterClaim
```

anandf:
I am interested to work on this bug. Can this be assigned to me.

PR #354, opened 2023-07-10
Using the right API version v1alpha1 for ClusterClaims

anandf:
Fixes https://github.com/open-cluster-management-io/clusteradm/issues/350

`ClusterClaim` API is still at version `cluster.open-cluster-management.io/v1alpha1`, however, the current code checks `ClusterClaim` resource in  version `cluster.open-cluster-management.io/v1` and does not find it, so it gives out an invalid error message saying that the command should be run against a managed cluster instead of an hub cluster. Using the right `apiVersion` fixes the issue.


xuezhaojun:
/assign @qiujian16 
I think this PR needs a approve to trigger the workflow because it's from a first-time contributor.

qiujian16:
thanks for your contribution!

PR #356, opened 2023-07-13
pick newest secret while running get bootstrap token

ycyaoxdu:
For now, if there are many bootstrap secret in `kube-system` ns, the `clusteradm get token --use-bootstrap-token` command would just fetch the last one in the list. This is not a excepted behavior because the last one is not the newest one.

### DIFF
--- a/pkg/cmd/get/token/exec.go
+++ b/pkg/cmd/get/token/exec.go
@@ -4,6 +4,7 @@ package token
 import (
 	"context"
 	"fmt"
+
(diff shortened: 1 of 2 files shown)


ycyaoxdu:
should we do some restriction on this situation or just pick the newest one? WDYT? @qiujian16

qiujian16:
seems pick the latest makes sense to me

qiujian16 (commented):
/approve
/lgtm

Issue #361, opened 2023-08-04
join and accept allow duplicate cluster-name


aii-nozomu-oki:
If I run `clusteradm join` with the same cluster-name from another cluster, the join succeeds and two CSRs are created in the hub cluster.
In this case, `clusteradm accept` will approve the two CSRs.

Also, if join and approve are done only once and then join and approve are done with the same cluster-name, the join will succeed as well.

The two clusters seem to watch the same namespace in the hub cluster.

Are these expected behaviors?

clusteradm version: v0.6.0

qiujian16:
hrm, I think we should at least print a warning, and let user to choose one csr to approve, instead of directly approve all the csr.

aii-nozomu-oki:
@qiujian16 
Thanks for your efforts.
Incidentally, the same problem exists in the ManagedClusterAutoApproval feature gate.

qiujian16:
yeah we have to resolve it in the controller also cc @skeeey

PR #369, opened 2023-09-05
🐛 Do not approve agent join with multiple csrs

qiujian16:
fixes https://github.com/open-cluster-management-io/clusteradm/issues/361


ohkinozomu:
This fix doesn't address the case where CSRs were already [garbage-collected](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#request-signing-process). 
Is it an intentional decision?

qiujian16:
I think garbage collected csr will be removed by the system. Let me double check if any state checking is missing

ohkinozomu:
I'm concerned with the following case:

1. A CSR is created with the name cluster-1 (CSR-1)
2. CSR-1 is approved
3. CSR-1 is garbage collected
4. An another CSR is created with the name cluster-1 (CSR-2)

In this case, the number of elements in `requesters` is 1, so the condition `if len(requesters) > 1` is skipped.
Is this correct?

qiujian16:
hrm this is a good catch.

> I'm concerned with the following case:
> 
> 1. A CSR is created with the name cluster-1 (CSR-1)
> 2. CSR-1 is approved
> 3. CSR-1 is garbage collected
> 4. An another CSR is created with the name cluster-1 (CSR-2)
> 
> In this case, the number of elements in `requesters` is 1, so the condition `if len(requesters) > 1` is skipped. Is this correct?

hrm this is a good catch. I have not considered this case.

PR #370, opened 2023-09-06
Add create addon cmd

qiujian16:


qiujian16:
/assign @zhujian7 
/hold 
still in progress

pkg/cmd/addon/create/cmd.go
```diff
@@ -0,0 +1,54 @@
...
+			}
+			if err := o.Run(); err != nil {
+				return err
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVarP(&o.Version, "name", "v", "0.0.1", "Specified version of the addon")
```
zhujian7, line 49:
```
clusteradm addon create --help

panic: unable to redefine 'v' shorthand in "create" flagset: it's already used for "name" flag
```

The `-v` here conflicts with the global option `-v`(number for the log level verbosity)

zhujian7:
```suggestion
	cmd.Flags().StringVar(&o.Version, "version", "0.0.1", "Specified version of the addon")
```

pkg/cmd/addon/create/exec.go
```diff
@@ -0,0 +1,225 @@
...
+		},
+	}
+
+	return addon, nil
+}
+
+func newClusterManagementAddon(o *Options) *addonv1alpha1.ClusterManagementAddOn {
+	cma := &addonv1alpha1.ClusterManagementAddOn{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: o.Name,
```
zhujian7, line 44: Should we add the annotation:
```
addon.open-cluster-management.io/lifecycle: "addon-manager"
```

pkg/cmd/addon/create/options.go
```diff
@@ -0,0 +1,37 @@
...
+	// version is the version of the addon
+	Version string
+
+	Overwrite bool
+
+	FileNameFlags genericclioptions.FileNameFlags
+	//
+	Streams genericclioptions.IOStreams
+
+	Placements []string
```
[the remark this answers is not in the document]
qiujian16: Rethinking it a bit, I do not think we should set installstartegy on create command, maybe a config cmd later would be better.

zhujian7:
@qiujian16  could you also add some e2e for this?

Issue #342, opened 2023-05-24
clusteradm addon enable does not do any validation or warning about unknown addons


nirs:
clusteradm addon manages specific addons, as documented in
https://open-cluster-management.io/getting-started/integration/

but will happily *enable* any name - for example:

```
$ clusteradm addon enable --names foo-bar --clusters dr1,dr2 --context hub
Deploying foo-bar add-on to namespaces open-cluster-management-agent-addon of managed cluster: dr1.
Deploying foo-bar add-on to namespaces open-cluster-management-agent-addon of managed cluster: dr2.
```

On the hub, you see that the new *addon* was added but it remain in `Unknown` state
since there is no such addon.

```
$ kubectl get managedclusteraddons -A --context hub | grep foo-bar
dr1         foo-bar                       Unknown                
dr2         foo-bar                       Unknown                
```

I think the user expectation is:
1. Either don't allow addons that are unknown
2. Or at least warn that this addon is unknown to alert users

See also the discussion in slack:
https://kubernetes.slack.com/archives/C01GE7YSUUF/p1684951933479419

mikeshng:
Maybe we need to check against the `ClusterManagementAddOn` first but I am not sure if `ClusterManagementAddOn` is required.

@ycyaoxdu what do you think?

qiujian16:
ClusterManagementAddOn is actually required

PR #374, opened 2023-09-20
return error for enabling an unknown addon

zhujian7:
Fix: https://github.com/open-cluster-management-io/clusteradm/issues/342

### DIFF
--- a/pkg/cmd/addon/enable/exec.go
+++ b/pkg/cmd/addon/enable/exec.go
@@ -107,6 +107,14 @@ func (o *Options) runWithClient(clusterClient clusterclientset.Interface,
 	}
 
 	for _, addon := range addons {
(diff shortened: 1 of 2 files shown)


pkg/cmd/addon/enable/exec.go
```diff
@@ -106,7 +106,22 @@ func (o *Options) runWithClient(clusterClient clusterclientset.Interface,
...
 	}
 
+	cmas, err := addonClient.AddonV1alpha1().ClusterManagementAddOns().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
 	for _, addon := range addons {
+		cmaExist := false
+		for _, cma := range cmas.Items {
```
qiujian16, line 116: can we just get cma here instead of list? It will require less permission.

zhujian7: PTAL

qiujian16 (commented):
/approve
/lgtm

PR #388, opened 2023-11-04
:bug: Disable the propagator webhook; Add ClusterManagementAddon manifests

dhaiducek:

## Summary

A GRC webhook was added, but it requires cert-manager as a prerequisite. Until we can resolve/update that, leave it disabled by default.

Also, there was a check added for ClusterManagementAddon, which were not deployed originally, currently causing an error during enabling app and policy addons:
https://github.com/open-cluster-management-io/clusteradm/blob/ed1069349e4f05faea934863035acedd690bbb2c/pkg/cmd/addon/enable/exec.go#L110-L114

/cc @mprahl @yiraeChristineKim @mikeshng @qiujian16 @ycyaoxdu @yue9944882

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ check-copyright:
 	@build/check-copyright.sh
 
 .PHONY: test
-test:
+test: deps
 	@build/run-unit-tests.sh
 
 .PHONY: clean-test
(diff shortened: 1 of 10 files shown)


dhaiducek:
Sorry for blowing up the reviewer list here, but I'm verifying `clusteradm` before KubeCon North America starting Monday and discovered enabling the app/policy addons didn't work at all. 😢 

If this all looks good, after it merges I'd appreciate if we could cut a new release on (or before) Monday. (My presentation is at 6 PM UTC on Monday.)

Issue #383, opened 2023-10-16
Re-init doesn't work well


aii-nozomu-oki:
**Describe the bug**
A second init with different options may not change the resource.

**To Reproduce**

Run init twice with different option

1st: `clusteradm init`

2nd: `clusteradm init --feature-gates=ManagedClusterAutoApproval=true`

The featureGates of the clustermanager object are as follows

```
$ kubectl get clustermanager cluster-manager -o yaml
...
  registrationConfiguration:
    featureGates:
    - feature: DefaultClusterSet
      mode: Enable
```

**Expected behavior**

The featureGates should be as follows

```
    featureGates:
    - feature: ManagedClusterAutoApproval
      mode: Enable
    - feature: DefaultClusterSet
      mode: Enable
```

**Environment ie: OCM version, clusteradm version, Kubernetes version and provider:**

OCM: v0.12.0
clusteradm: latest(2d580956d97cbbbcfdad8dea4c5456f742ecce7c)
Kubernetes: K3s v1.27.1+k3s1

**Additional context**

Not sure if re-init is supported, but wanted to try the init again to change the feature gate.

qiujian16:
hrm I think it should be supported.

aii-nozomu-oki:
Thanks for the fix!

PR #389, opened 2023-11-06
🐛 Fix a bug in apply resources

qiujian16:
This may impact multiple functions

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/clusteradm/issues/383

### DIFF
--- a/pkg/helpers/reader/reader.go
+++ b/pkg/helpers/reader/reader.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"embed"
 	"fmt"
+
 	"github.com/jonboulle/clockwork"
 	"github.com/openshift/library-go/pkg/assets"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/helpers/reader/reader.go
```diff
@@ -88,6 +89,11 @@ func (r *ResourceReader) applyOneObject(info *resource.Info) error {
 	helper := resource.NewHelper(info.Client, info.Mapping).
 		DryRun(r.dryRun)
 
+	modified, err := kubectlutil.GetModifiedConfiguration(info.Object, false, unstructured.UnstructuredJSONScheme)
```
zhujian7, line 92: Hi @qiujian16 could you add some comments/tests for this, I do not understand why moving this func upper can fix the issue.

qiujian16: this is to get the modified object, it has to be called before info.Get() otherwise the info.Object will be from the server. A test is added.

zhujian7:
/lgtm
/hold
@qiujian16 feel free to unhold

PR #393, opened 2023-11-09
Deploy webhook config for subadmin workarounds

dhaiducek:
Without the webhook, the subscription operator would not detect subscription admins.

## Summary

The subscription operator detects whether a user is a subscription admin based on the `open-cluster-management.io/user-identity` and `open-cluster-management.io/user-group` injected by a webhook. Adding the webhook to `clusteradm` will give one less step for the user to use this workaround, where they only need to add the annotations to the subscription to deploy the application.

/cc @xiangjingli @mikeshng


mikeshng (approved):
/approve

/lgtm

Thank you Dale!

PR #398, opened 2023-12-02
:sparkles: clusteradm join support resource-qos-class parameter

haowells:

## Summary
1. Klusterlet cr add new flag UseBestEffortResource which allow user use besteffort setting for deployment agent, registration, work etc
related PR https://github.com/open-cluster-management-io/ocm/pull/317
2. Add bool flag "createNamespace" to allow use existing open-cluster-management-* ns. reason behind it because our k8s setup have bunch of customized ns label/annotations that is different from default
## Related issue(s)

Fixes #


qiujian16 (commented):
I think you would need to update the api library in go.mod and also the manifests under join cmd.

haowells:
> I think you would need to update the api library in go.mod and also the manifests under join cmd.

@qiujian16 update api to latest commit and push new commit
go get open-cluster-management.io/api@97a8a92a7f30a0aaf123b8a7a9fa94859baea15b

haowells:
> there is an error during tests
> 
> ```
> Error: error when creating "local": Klusterlet.operator.open-cluster-management.io "klusterlet" is invalid: spec.resourceRequirement.type: Unsupported value: "default": supported values: "Default", "BestEffort"
> ```
address in this commit https://github.com/open-cluster-management-io/clusteradm/pull/398/commits/daea1057a93940837c3df7c2fe1d29e54847ba34

PR #403, opened 2024-02-29
🌱 Update default version to 0.13.0

qiujian16:

## Related issue(s)

Fixes #


pkg/helpers/version/version.go
```diff
@@ -44,66 +44,6 @@ func GetVersionBundle(version string) (VersionBundle, error) {
 
 	// predefined bundle version
 	// TODO: automated version tracking
-	versionBundleList["0.5.0"] = VersionBundle{
```
zhujian7, line 47: so this means we only support n-3 versions?

pkg/helpers/version/version.go
```diff
@@ -136,6 +76,17 @@ func GetVersionBundle(version string) (VersionBundle, error) {
...
 	}
 
+	versionBundleList["0.13.0"] = VersionBundle{
+		Registration:             "v0.13.0",
+		Placement:                "v0.13.0",
+		Work:                     "v0.13.0",
+		Operator:                 "v0.13.0",
+		AddonManager:             "v0.13.0",
+		AppAddon:                 "v0.13.0",
+		PolicyAddon:              "v0.13.0",
```
gparvin, line 86: I don't see updated versions for any of the components (not the ones I checked).  Is the 0.13.0 release in the works now?

pkg/helpers/version/version.go
qiujian16: I think so, we can not support earlier version since the change on operator manifests.

pkg/helpers/version/version.go
qiujian16: it is in the progressing Operator/AddonManager have 0.13.0 already.

VERSION.txt
```diff
@@ -1 +1 @@
-v0.7.2
+v0.8.0
```
zhujian7, line 1: Is this file necessary? can we remove it?

qiujian16: let's add a todo...

dhaiducek: It's referenced here for CLI versioning:
https://github.com/open-cluster-management-io/clusteradm/blob/bb97f150a2340e6cfe7d407c7a71af8c39eaeaf3/version.go#L9

PR #408, opened 2024-03-29
:seedling: Update integration tests

dhaiducek:

## Summary

- ~Add a pre-flight test run in the release workflow that sets the bundle version for the integration tests (and could be used in other tests as well)~
- Use the `ginkgo` CLI to run tests (I find the output a bit cleaner since our tests are written with Ginkgo)
- Refactors the addon integration tests, making it a bit easier to maintain and also add an addon in the future, also adding a deployment readiness check.

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,12 +4,12 @@ on:
   push:
     # Sequence of patterns matched against refs/tags
     tags:
-      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+      - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
 
 name: Create Release
 
 env:
-  GO_VERSION: '1.21'
+  GO_VERSION: "1.21"
 
 jobs:
   build:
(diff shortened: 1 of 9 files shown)


qiujian16 (commented):
/approve

/assign @zhujian7

.github/workflows/release.yaml
```diff
@@ -25,6 +25,14 @@ jobs:
       - name: Check version
         run: |
           cat VERSION.txt | grep ${{ github.ref_name }}
+      - name: Preflight test
+        run: |
+          echo "::group::E2E Tests"
+          make test-e2e
+          echo "::endgroup::"
+          echo "::group::Integration Tests"
+          OCM_BUNDLE_VERSION=$(cat VERSION.txt) make test-integration
```
zhujian7, line 34: when I try to run `OCM_BUNDLE_VERSION=$(cat VERSION.txt) make test-integration` locally, I got some errors `couldn't find the requested version bundle: 0.8.0`

dhaiducek: Ah, you're right--I'll take this out. I wasn't thinking about the VERSION being clusteradm and not OCM. Since this uses the "default" version now these tests would run in the PR when the OCM version is updated, so I'm not so sure a preflight check is even necessary. We'll add some additional E2E tests to cover the managed cluster add-on availability in the future.

dhaiducek: Updated

zhujian7:
LGTM

Issue #410, opened 2024-04-04
clusteradm join --wait does not wait until the cluster is joined


nirs:
**Describe the bug**

`clusteradm join --wait` can return before the managedcluster/{clustername} resource was created on the hub.

**To Reproduce**

1. Initialize the hub:
   ```   
   clusteradm init --featue-gates ManagedClusterAutoApproval=true --wait --context hub
   ```
2. Install hub-addons
   ```
   clusteradm install hub-addon --names application-manager,governance-policy-framework --context hub
   ```

3. Wait until all the deployments are rolled out
   ```
   for d in cluster-manager governance-policy-addon-controller governance-policy-propagator multicluster-operators-appsub-summary multicluster-operators-channel multicluster-operators-placementrule multicluster-operators-subscription; do
       kubectl rollout status deploy/$d -n open-cluster-management --timeout 300s --context hub
   done
   for d in cluster-manager-placement-controller cluster-manager-registration-controller cluster-manager-registration-webhook cluster-manager-work-webhook; do
      kubectl rollout status deploy/$d -n open-cluster-management-hub --timeout 300s --context hub
   done
   ```
4. Join a managed cluster
   ```
   clusteradm join --hub-token ... --hub-apiserver ... --cluster-name dr1 --wait --context dr1
   ```
5. Wait until the managed cluster is accepted on the hub
   ```
   kubectl wait managedcluster/dr1 --for=jsonpath='{.spec.hubAcceptsClient}=true' --timeout 60s --context hub
   ```
The command randomly fail with:

    Error from server (NotFound): managedclusters.cluster.open-cluster-management.io "dr1" not found

**Expected behavior**

The purpose of the `--wait` option is to help users to use the tool without caring about implementation details that may change. Thus the command should return only when the cluster was joined, not when clusteradm sent a request to the agent on the managed cluster.

When the command returns, the user should be able to enable addons on the managed cluster.

**Environment ie: OCM version, clusteradm version, Kubernetes version and provider:**

- clusteradm from be5413067cf82ff29735c7244ca16281ba8f7459
- ocm images built from (https://github.com/open-cluster-management-io/ocm/pull/388/commits/2c8c0872bfaddea95890d6486e2daedbb5f5eb66

**Additional context**

This is easy to work around by waiting until the managedclsuter resource is created on the hub.

qiujian16:
[...]

PR #411, opened 2024-04-08
🐛 Check if managed cluster is created during join

qiujian16:
This is to check if managed cluster is created
when using --wait flag with join command, using
the bootstrap secret

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/clusteradm/issues/410

### DIFF
--- a/pkg/cmd/join/exec.go
+++ b/pkg/cmd/join/exec.go
@@ -29,7 +29,9 @@ import (
 	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/cmd/util"
+	clusterclient "open-cluster-management.io/api/client/cluster/clientset/versioned"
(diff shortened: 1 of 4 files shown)


mikeshng:
CC @nirs

pkg/cmd/join/exec.go
```diff
@@ -455,6 +462,51 @@ func checkIfRegistrationOperatorAvailable(f util.Factory) (bool, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to create rest config: %v", err)
+	}
+	clusterClient, err := clusterclient.NewForConfig(restConfig)
+	if err != nil {
+		return fmt.Errorf("failed to create client: %v", err)
+	}
+
+	phase := &atomic.Value{}
+	phase.Store("")
```
zhujian7, line 478: what is the `phase.Store("")` used for?
we did not print any status

PR #412, opened 2024-04-08
🌱 Change the default version to v0.13.1

qiujian16:

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -46,7 +46,7 @@ type VersionBundle struct {
 	MulticlusterControlplane string
 }
 
-var defaultBundleVersion = "0.13.0"
+var defaultBundleVersion = "0.13.1"
 
 func GetDefaultBundleVersion() string {
 	return defaultBundleVersion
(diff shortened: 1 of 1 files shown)


pkg/version/version.go
```diff
@@ -116,6 +116,17 @@ func GetVersionBundle(version string) (VersionBundle, error) {
 		MulticlusterControlplane: "v0.4.0",
 	}
 
+	versionBundleList["0.13.1"] = VersionBundle{
```
zhujian7, line 119: Should we consider changing the bundle version from x.y.z to x.y? so we only update the versions of the components when there are any z stream updates?

zhujian7: for example:
```
        var defaultBundleVersion = "0.13"

	versionBundleList["0.13"] = VersionBundle{
		Registration:             "v0.13.1",
		Placement:                "v0.13.1",
		Work:                     "v0.13.1",
		Operator:                 "v0.13.1",
		AddonManager:             "v0.13.1",
		AppAddon:                 "v0.13.0",
		PolicyAddon:              "v0.13.0",
		MulticlusterControlplane: "v0.4.0",
	}
```

qiujian16: we cannot currently I think. Bundle version is an input value user can specify, and we have an upgrade cmd to allow upgrade from one zstream to another. If we change bundle version to only y stream, it will be strange to run upgrade for z streams.

Issue #416, opened 2024-04-19
`clusteradm join --dry-run` error and output missing manifests


EricaJ6:
**Describe the bug**
Before we actually join the managed cluster to the hub, we would like to preview the resources that will be created in the managed cluster.
`clusteradm join --dry-run=true` produces below error
`Error: resource mapping not found for name: "klusterlet" namespace: "" from "local": no matches for kind "Klusterlet" in version "operator.open-cluster-management.io/v1"
ensure CRDs are installed first`
Also, with an additional `--output-file` flag, it outputs to the terminal instead of the file.

Without the `--dry-run=true` flag, it's successful without any errors.
While comparing the manifests with or without the `--dry-run` flag, with `--dry-run` flag one is missing 

```
# Copyright Contributors to the Open Cluster Management project
apiVersion: operator.open-cluster-management.io/v1
kind: Klusterlet
metadata:
  name: klusterlet
spec:
  deployOption:
    mode: Singleton
  registrationImagePullSpec: quay.io/open-cluster-management/registration:v0.13.1
  workImagePullSpec: quay.io/open-cluster-management/work:v0.13.1
  imagePullSpec: quay.io/open-cluster-management/registration-operator:v0.13.1
  clusterName: m1
  namespace: open-cluster-management-agent 
  externalServerURLs:
  
  - url: https://m1-control-plane:6443
  
  resourceRequirement:
    type: Default
  
  
    # Uncomment the following configuration lines to add hostAliases for hub api server, 
    # if the server field in your hub cluster kubeconfig is a domain name instead of an ipv4 address.
    # For example, https://xxx.yyy.zzz.
    # This configuration is used for registration-agent and work-agent to communicate with hub api server.
    # hubApiServerHostAlias:
    # ip: "1.2.3.4"
    # hostname: "xxx.yyy.zzz"

---
```

**To Reproduce**
1. `clusteradm join --hub-token $token --hub-apiserver $apiserver --cluster-name $managed_cluster_name --singleton --dry-run=true --output-file='manifests.yaml'
`
2. `clusteradm join --hub-token $token --hub-apiserver $apiserver --cluster-name $managed_cluster_name --singleton --dry-run=true
`
3. `clusteradm join --hub-token $token --hub-apiserver $apiserver --cluster-name $managed_cluster_name --singleton --output-file='manifests.yaml'
`

**Expected behavior**
With step 1, it outputs the full manifests (including Klusterlet) to the file without any error.
[...]

PR #421, opened 2024-04-25
🐛 Ignore CRD not installed error when dry-run

zhujian7:

## Related issue(s)
https://github.com/open-cluster-management-io/clusteradm/issues/416

Fixes #

### DIFF
--- a/pkg/helpers/reader/reader.go
+++ b/pkg/helpers/reader/reader.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"embed"
 	"fmt"
+	"strings"
 
 	"github.com/jonboulle/clockwork"
 	"github.com/openshift/library-go/pkg/assets"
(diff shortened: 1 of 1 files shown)


pkg/helpers/reader/reader.go
```diff
@@ -60,6 +61,12 @@ func (r *ResourceReader) Apply(fs embed.FS, config interface{}, files ...string)
 		Stream(bytes.NewReader(rawObjects), "local").
 		Flatten().
 		Do()
+	if r.dryRun {
+		// Ignore CRD not installed error when dry-run
+		rb.IgnoreErrors(func(err error) bool {
+			return err != nil && strings.Contains(err.Error(), "ensure CRDs are installed first")
```
qiujian16, line 67: just curious, do we have a certain error type for this?

zhujian7: seems no

PR #428, opened 2024-05-28
✨ Add capi into join flow

qiujian16:

## Summary

so we can use
```
clusteradm join .... --capi-import --capi-cluster-name aws-cluster
```
to add a capi cluster to ocm
## Related issue(s)

Fixes #


qiujian16:
/assign @zhujian7
I did some test locally already

zhujian7:
/lgtm
/hold
@qiujian16 should we create a new release 0.8.3 to include the repo chart URL fix before this merge?

qiujian16:
yeah, good point.

Issue #433, opened 2024-06-03
warning event of FailedToRetrieveImagePullSecret


zhiweiyin318:
**Describe the bug**
after a cluster is imported to the Hub cluster ,there is warning event shows 

```
open-cluster-management-agent         72s (x4 over 78s)       Warning   FailedToRetrieveImagePullSecret    Pod/klusterlet-registration-agent-6477d4d57f-jfxbp                                 Unable to retrieve some image pull secrets (open-cluster-management-image-pull-credentials); attempting to pull the image may not succeed.
```

**To Reproduce**

**Expected behavior**

**Environment ie: OCM version, clusteradm version, Kubernetes version and provider:**

**Additional context**

zhiweiyin318:
the PR for clusteradm is merged in which it will create an empty image pull secret by default when run `clusteradm init` or `clusteradm join`, and also add a flag `image-pull-credential-file` to `clusteradm init/join`  to config image pull secret for users.

PR #438, opened 2024-06-13
:bug:  create empty pull secret in the operator ns

zhiweiyin318:

## Related issue(s)

https://github.com/open-cluster-management-io/clusteradm/issues/433

Fixes #


pkg/cmd/init/exec.go
```diff
@@ -219,7 +219,13 @@ func (o *Options) run() error {
 			}
 		}
 
-		//if service-account wait for the sa secret
+		//  apply an empty image pull secret if the image pull secret is not found in the ns.
+		err = helpers.ApplyImagePullSecret(context.TODO(), "open-cluster-management", kubeClient)
```
zhujian7, line 223: if it is dry run, will we create the secret?

pkg/cmd/join/exec.go
```diff
@@ -330,6 +330,12 @@ func (o *Options) run() error {
 		return err
 	}
 
+	//  apply an empty image pull secret if the image pull secret is not found in the ns.
+	err = helpers.ApplyImagePullSecret(context.TODO(), "open-cluster-management", kubeClient)
```
zhujian7, line 334: please also consider dry-run, and I am not sure if this should be put into the above `applyKlusterlet` func, is this secret used by the klusterlet-operator? if it is, I think we should put it into the `applyKlusterlet` func, and use the constant `OperatorNamesapce` to replace the "open-cluster-management"

pkg/cmd/init/exec.go
zhiweiyin318: I think we need to create the secret . have fixed.

pkg/cmd/join/exec.go
zhiweiyin318: There is no image pull secret in the operator deployment or sa. so no need to consider the order.

pkg/cmd/init/scenario/init/image-pull-secret.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: open-cluster-management-image-pull-credentials
+  namespace: open-cluster-management
+  labels:
+    app: cluster-manager
+type: kubernetes.io/dockerconfigjson
+data:
+  .dockerconfigjson: e30K  # "{}" empty value here
```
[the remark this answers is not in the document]
zhujian7: oh, it is secret, so should be base64ed.

qiujian16 (commented):
ideally, should have a flag to input docker creds, but we could do it in the future.

pkg/helpers/client.go
```diff
@@ -38,6 +38,8 @@ const (
 	BootstrapToken      TokenType = "bootstrap-token"
 	ServiceAccountToken TokenType = "service-account-token"
 	UnknownToken        TokenType = "unknown-token"
+
+	ImagePullSecret = "open-cluster-management-image-pull-credentials"
```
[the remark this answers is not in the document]
zhiweiyin318: this is used in e2e test, let me move it to e2e part.

pkg/cmd/init/cmd.go
```diff
@@ -59,10 +59,12 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	cmd.Flags().StringToStringVar(&o.resourceRequests, "resource-requests", nil, "the resource requests of all the containers managed by the cluster manager and the cluster manager operator, for example: cpu=500m,memory=500Mi")
 	cmd.Flags().BoolVar(&o.createNamespace, "create-namespace", true, "If true, create open-cluster-management namespace, otherwise use existing one")
 
-	//clusterManagetSet contains the flags for deploy cluster-manager
+	// clusterManagetSet contains the flags for deploy cluster-manager
 	clusterManagerSet := pflag.NewFlagSet("clusterManagerSet", pflag.ExitOnError)
 	cmd.Flags().StringVar(&o.registry, "image-registry", "quay.io/open-cluster-management",
 		"The name of the image registry serving OCM images, which will be applied to all the deploying OCM components.")
+	cmd.Flags().StringVar(&o.imagePullCred, "image-pull-credential", "",
```
qiujian16, line 66: actually, I think we should set a file instead of a base64 encoded string, it is not user friendly.

zhiweiyin318:
@qiujian16 @zhujian7  please take another look.
1. changed to `image-pull-credential-file`.
2. add the pull secret to operator deployments too since the user can configure it via flag.

PR #441, opened 2024-06-24
🌱 update deps to v0.14

qiujian16:

## Related issue(s)

Fixes #


qiujian16:
@dhaiducek @mikeshng would you check if anything needs to be updated for policy and app addon?

pkg/version/version.go
```diff
@@ -149,6 +128,17 @@ func GetVersionBundle(version string) (VersionBundle, error) {
...
 	}
 
+	versionBundleList["0.14.0"] = VersionBundle{
+		Registration:             "0.14.0",
+		Placement:                "0.14.0",
+		Work:                     "0.14.0",
+		Operator:                 "0.14.0",
+		AddonManager:             "0.14.0",
+		AppAddon:                 "0.14.0",
+		PolicyAddon:              "0.14.0",
```
dhaiducek, line 138: Why are there no `v` prefixes here? Is that handled elsewhere?

dhaiducek (changes requested):
I've just released v0.14.0 for Policy, so we're good there. I have one question about the lack of `v` prefix in this PR.

mikeshng (approved):
/lgtm

Good catch by Dale. Otherwise, all good to go.

pkg/version/version.go
qiujian16: updated

mikeshng (approved):
/lgtm

dhaiducek:
Also, the default version here is still set to `v1.13.3`. That will need to be updated before release.

Issue #442, opened 2024-06-24
Add --client-cert-expiration-seconds option


ohkinozomu:
**Describe the enhancement**
Currently, clusteradm does not set the `spec.expirationSeconds` for CSRs.
As a result, the default value is used, which can lead to certificates being issued with excessively long expiration periods.
For example, in GKE standard, this period is 5 years (I couldn't find documentation on this).

I propose adding the `--client-cert-expiration-seconds` option to set spec.expirationSeconds.

PR #443, opened 2024-06-24
✨ Add --client-cert-expiration-seconds

ohkinozomu:

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/clusteradm/issues/442

### DIFF
--- a/pkg/cmd/join/cmd.go
+++ b/pkg/cmd/join/cmd.go
@@ -76,5 +76,6 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	cmd.Flags().StringToStringVar(&o.resourceRequests, "resource-requests", nil, "the resource requests of all the containers managed by the klusterlet and the klusterlet operator, for example: cpu=500m,memory=500Mi")
 	cmd.Flags().BoolVar(&o.createNameSpace, "create-namespace", true, "If true, create the operator namespace(open-cluster-management) and the agent namespace(open-cluster-management-agent for Default mode, <klusterlet-name> for Hosted mode), otherwise use existing one")
 	cmd.Flags().BoolVar(&o.enableSyncLabels, "enable-sync-labels", false, "If true, sync the labels from klusterlet to all agent resources.")
+	cmd.Flags().Int32Var(&o.clientCertExpirationSeconds, "client-cert-expiration-seconds", 31536000, "clientCertExpirationSeconds represents the seconds of a client certificate to expire.")
 	return cmd
 }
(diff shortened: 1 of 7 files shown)


pkg/cmd/join/cmd.go
```diff
@@ -76,5 +76,6 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	cmd.Flags().StringToStringVar(&o.resourceRequests, "resource-requests", nil, "the resource requests of all the containers managed by the klusterlet and the klusterlet operator, for example: cpu=500m,memory=500Mi")
 	cmd.Flags().BoolVar(&o.createNameSpace, "create-namespace", true, "If true, create the operator namespace(open-cluster-management) and the agent namespace(open-cluster-management-agent for Default mode, <klusterlet-name> for Hosted mode), otherwise use existing one")
 	cmd.Flags().BoolVar(&o.enableSyncLabels, "enable-sync-labels", false, "If true, sync the labels from klusterlet to all agent resources.")
+	cmd.Flags().Int32Var(&o.clientCertExpirationSeconds, "client-cert-expiration-seconds", 0, "clientCertExpirationSeconds represents the seconds of a client certificate to expire. If it is not set or 0, the default duration seconds will be set by the hub cluster.")
```
qiujian16, line 79: should it not to be default 0?

ohkinozomu: I have prioritized compatibility with existing behavior, but it might be good to set a default value.

I can't think of an appropriate default value, but it might be worth considering that [the default for Kubernetes' ClusterSigningDuration is one year](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#signers).
What do you think?

qiujian16: yes, that makes sense

pkg/cmd/join/cmd.go
```diff
@@ -76,5 +76,6 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	cmd.Flags().StringToStringVar(&o.resourceRequests, "resource-requests", nil, "the resource requests of all the containers managed by the klusterlet and the klusterlet operator, for example: cpu=500m,memory=500Mi")
 	cmd.Flags().BoolVar(&o.createNameSpace, "create-namespace", true, "If true, create the operator namespace(open-cluster-management) and the agent namespace(open-cluster-management-agent for Default mode, <klusterlet-name> for Hosted mode), otherwise use existing one")
 	cmd.Flags().BoolVar(&o.enableSyncLabels, "enable-sync-labels", false, "If true, sync the labels from klusterlet to all agent resources.")
+	cmd.Flags().Int32Var(&o.clientCertExpirationSeconds, "client-cert-expiration-seconds", 31536000, "clientCertExpirationSeconds represents the seconds of a client certificate to expire. If it is not set or 0, the default duration seconds will be set by the hub cluster.")
```
qiujian16, line 79: the description should be changed a bit?

PR #444, opened 2024-06-25
:bug: make the clusteradm join output more clear

### LINKED ISSUE #372: Support join cluster with auto-approve
We have support 
> ManagedClusterAutoApproval will approve a managed cluster registraion request automatically. A registraion request to be automatically approved must be initiated by specific users, these users are specifed with "--cluster-auto-approval-users" flag in the registration hub controller.

https://github.com/open-cluster-management-io/api/blob/ad02e8269c0919e4fc4e546b3e08fb3468130748/feature/feature.go#L61

Even if the cluster enables this feature, when use `clusteradm join ...`, it always asks the user to run `cluseradm accept ...`

The expected behaviour is we can pass "--cluster-auto-approval-user" to `clusteradm join` command so that the CSR is created by this user and can be approved automatically. we do not need to prompt run `clusteradm accept` anymore.

FYI @qiujian16 @skeeey

skeeey:

## Related issue(s)

Fixes #372

### DIFF
--- a/pkg/cmd/join/exec.go
+++ b/pkg/cmd/join/exec.go
@@ -357,6 +357,7 @@ func (o *Options) run() error {
 
 	fmt.Fprintf(o.Streams.Out, "Please log onto the hub cluster and run the following command:\n\n"+
 		"    %s accept --clusters %s\n\n", helpers.GetExampleHeader(), o.values.ClusterName)
+	fmt.Fprintf(o.Streams.Out, "This is not needed when the ManagedClusterAutoApproval feature is enabled\n")
 	return nil
 
 }


pkg/cmd/init/scenario/init/bootstrap_cluster_role.yaml
```diff
@@ -34,4 +34,9 @@ rules:
   - "managedclustersets/join"
   verbs: 
   - "create"
-  
\ No newline at end of file
+- apiGroups:
```
qiujian16, line 37: I do no think agent needs this clusterrole?

skeeey: the auto approval feature is controlled by the clustermanager, when we use `clusteradm join` a cluster, we need read the clustermanager to see it if enable the auto approval feature, so I added  this clusterrole for bootstrap token

qiujian16: But it is mainly for clusteradm to know, not agent, right?

skeeey: right, it is mainly for clusteradm to know

skeeey: for clusteradm join, it seems we only have the bootstrap token, right?

skeeey: or we give a clear messages instead of checking the auto approval feature? e.g.

```
Your managed cluster cluster1 has joined the Hub.

If the ManagedClusterAutoApproval feature is not enabled, Please log onto the hub ...
```

qiujian16: yes, I think this would be better.

pkg/cmd/join/exec.go
```diff
@@ -355,7 +355,8 @@ func (o *Options) run() error {
 		}
 	}
 
-	fmt.Fprintf(o.Streams.Out, "Please log onto the hub cluster and run the following command:\n\n"+
+	fmt.Fprintf(o.Streams.Out, "\nYour managed cluster %s has joined the Hub.\n", o.values.ClusterName)
```
qiujian16, line 358: since  ManagedClusterAutoApproval  is disable by default, we might say
Please log onto the hub cluster and run the following command: at first and then say it is not needed when If the ManagedClusterAutoApproval feature is not enabled.

skeeey: updated

PR #465, opened 2025-01-15
:sparkles: Initialize hub with aws irsa

jaswalkiranavtar:

## Related issue(s)

Fixes # https://github.com/open-cluster-management-io/ocm/issues/514


pkg/cmd/init/exec.go
```diff
@@ -353,3 +359,24 @@ func (o *Options) deploySingletonControlplane(kubeClient kubernetes.Interface) e
...
+	if slices.Contains(o.registrationAuth, "csr") {
+		registrationDrivers = append(registrationDrivers, operatorv1.RegistrationDriverHub{AuthType: "csr"})
+	}
+	if slices.Contains(o.registrationAuth, "awsirsa") {
+		rawConfig, err := o.ClusteradmFlags.KubectlFactory.ToRawKubeConfigLoader().RawConfig()
+		if err != nil {
+			klog.Errorf("unable to load hub cluster kubeconfig: %v", err)
+			return nil, err
+		}
+		hubClusterArn := rawConfig.Contexts[rawConfig.CurrentContext].Cluster
```
qiujian16, line 374: why the arc is the cluster name? is it alway true on AKS?

jaswalkiranavtar:
Merge this PR first: https://github.com/open-cluster-management-io/ocm/pull/810

And update the go.mod which latest commit sha of ocm and then merge this.

pkg/cmd/init/exec.go
jaswalkiranavtar: Yes

jaswalkiranavtar: Adding another option flag called --hub-cluster-arn which can be used in tests as well as in case the hubclusterArn is missing in kubeconfig.

pkg/cmd/init/cmd.go
```diff
@@ -78,6 +81,9 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	_ = clusterManagerSet.SetAnnotation("singleton-name", "singletonSet", []string{})
 	o.Helm.AddFlags(singletonSet)
 	cmd.Flags().AddFlagSet(singletonSet)
+	cmd.Flags().StringArrayVar(&o.registrationAuth, "registration-auth", []string{}, "The type of authentication to use for registering and authenticating with hub, this flag can be repeated to specify multiple authentication types.")
+	cmd.Flags().StringVar(&o.hubClusterArn, "hub-cluster-arn", "",
```
qiujian16, line 85: I think you would need to validate this value

pkg/cmd/init/cmd.go
```diff
@@ -78,6 +81,9 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	_ = clusterManagerSet.SetAnnotation("singleton-name", "singletonSet", []string{})
 	o.Helm.AddFlags(singletonSet)
 	cmd.Flags().AddFlagSet(singletonSet)
+	cmd.Flags().StringArrayVar(&o.registrationAuth, "registration-auth", []string{}, "The type of authentication to use for registering and authenticating with hub, this flag can be repeated to specify multiple authentication types.")
```
qiujian16, line 84: it is better to validate the flag, only valid auth drvier is allowed. In the message, also should give the allowed value.

pkg/cmd/init/exec.go
```diff
@@ -353,3 +359,27 @@ func (o *Options) deploySingletonControlplane(kubeClient kubernetes.Interface) e
 	}
 	return nil
 }
+
+func getRegistrationDrivers(o *Options) ([]operatorv1.RegistrationDriverHub, error) {
```
qiujian16, line 363: could we make it simpler? we should validate and return error when a driver name is not allowed in advance.

```go
for _, driver := range o.registrationAuth:
    registrationDriver :=  operatorv1.RegistrationDriverHub{AuthType: driver}
   if driver == "awisra" {
      ...
   }
   registrationDrivers = append(registrationDrivers, registrationDriver)
```

pkg/cmd/init/cmd.go
jaswalkiranavtar: Thr validation for this is already added to cluster-manager spec and the validation error is already propagated back to the clusteradm consumer even now. 

Is that sufficient?

pkg/cmd/init/cmd.go
jaswalkiranavtar: Done

pkg/cmd/init/exec.go
jaswalkiranavtar: Done

jaswalkiranavtar:
Also commented the e2e test with a TODO

PR #468, opened 2025-01-31
✨ Add changes to clusteradm accept to disable csr update based on annotation on ManagedCluster 

jeffw17:

## Summary

For AWSIRSA, we want to disable CSR update when running clusteradm accept command and we verify by checking the annotation on the managed cluster to see if it contains the ARN. 

## Related issue(s)

Fixes # [514](https://github.com/open-cluster-management-io/ocm/issues/514)

### DIFF
--- a/pkg/cmd/accept/exec.go
+++ b/pkg/cmd/accept/exec.go
@@ -100,6 +100,9 @@ func (o *Options) accept(kubeClient *kubernetes.Clientset, clusterClient *cluste
 	if err != nil {
 		return false, fmt.Errorf("fail to get managedcluster %s: %v", clusterName, err)
 	}
+	// when a managed cluster registers with hub using awsirsa registration-auth, it will add this annotation
+	// to ManagedCluster resource, presense of which is used to decide the requested authentication type.
+	// awrirsa authentication doesn't create CSR on hub, hence there is nothing to approve
 	_, hasEksArn := managedCluster.Annotations["agent.open-cluster-management.io/managed-cluster-arn"]
 
 	var approved bool
--- a/pkg/cmd/accept/exec.go
+++ b/pkg/cmd/accept/exec.go
@@ -94,10 +94,24 @@ func (o *Options) runWithClient(kubeClient *kubernetes.Clientset, clusterClient 
 }
 
 func (o *Options) accept(kubeClient *kubernetes.Clientset, clusterClient *clusterclientset.Clientset, clusterName string, waitMode bool) (bool, error) {
-	approved, err := o.approveCSR(kubeClient, clusterName, waitMode)
+	managedCluster, err := clusterClient.ClusterV1().ManagedClusters().Get(context.TODO(),
+		clusterName,
+		metav1.GetOptions{})
 	if err != nil {
-		return approved, fmt.Errorf("fail to approve the csr for cluster %s: %v", clusterName, err)
+		return false, fmt.Errorf("fail to get managedcluster %s: %v", clusterName, err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/cmd/accept/exec.go
```diff
@@ -94,10 +94,24 @@ func (o *Options) runWithClient(kubeClient *kubernetes.Clientset, clusterClient
 }
 
 func (o *Options) accept(kubeClient *kubernetes.Clientset, clusterClient *clusterclientset.Clientset, clusterName string, waitMode bool) (bool, error) {
-	approved, err := o.approveCSR(kubeClient, clusterName, waitMode)
+	managedCluster, err := clusterClient.ClusterV1().ManagedClusters().Get(context.TODO(),
+		clusterName,
+		metav1.GetOptions{})
 	if err != nil {
-		return approved, fmt.Errorf("fail to approve the csr for cluster %s: %v", clusterName, err)
+		return false, fmt.Errorf("fail to get managed cluster")
```
mikeshng, line 101:
```suggestion
		return false, fmt.Errorf("fail to get managedcluster %s: %v", clusterName, err")
```
more consistent with the previous error message.

pkg/cmd/accept/exec.go
```diff
@@ -94,10 +94,24 @@ func (o *Options) runWithClient(kubeClient *kubernetes.Clientset, clusterClient
...
+	_, hasEksArn := managedCluster.Annotations["agent.open-cluster-management.io/managed-cluster-arn"]
+
+	var approved bool
+	if !hasEksArn {
+		approved, err = o.approveCSR(kubeClient, clusterName, waitMode)
+		if err != nil {
+			return approved, fmt.Errorf("fail to approve the csr for cluster %s: %v", clusterName, err)
+		}
+	} else {
+		approved = hasEksArn
```
mikeshng, line 112:
```suggestion
		approved = true
```
minor nit for readability.

mikeshng (commented):
I've added some nit comments, but my major concern is that it looks like anyone can just stamp the annotation on their managedcluster and bypass the CSR check. WDYT?

@jaswalkiranavtar Did we discuss this already and this is not a concern?

pkg/cmd/accept/exec.go
jaswalkiranavtar: Yes, this is part of our design. This will work for only EKS based registration. If somebody is using default csr authentication, and they add these annotations, their registration won't complete. If they want to complete the registration with csr based authentication, they have to remove this annotation.

pkg/cmd/accept/exec.go
```diff
@@ -94,10 +94,24 @@ func (o *Options) runWithClient(kubeClient *kubernetes.Clientset, clusterClient
...
 func (o *Options) accept(kubeClient *kubernetes.Clientset, clusterClient *clusterclientset.Clientset, clusterName string, waitMode bool) (bool, error) {
-	approved, err := o.approveCSR(kubeClient, clusterName, waitMode)
+	managedCluster, err := clusterClient.ClusterV1().ManagedClusters().Get(context.TODO(),
+		clusterName,
+		metav1.GetOptions{})
 	if err != nil {
-		return approved, fmt.Errorf("fail to approve the csr for cluster %s: %v", clusterName, err)
+		return false, fmt.Errorf("fail to get managedcluster %s: %v", clusterName, err)
+	}
+	_, hasEksArn := managedCluster.Annotations["agent.open-cluster-management.io/managed-cluster-arn"]
```
qiujian16, line 103: please add a comment here.

jaswalkiranavtar: Done. Does it look ok? Will sign the commit tomorrow.

PR #488, opened 2025-05-26
:sparkles: add `--bundle-version-overrides` flag; update default bundle to OCM v0.16.1

TylerGillson:

## Summary

- Add `--bundle-version-overrides` flag for customizing the version bundle used by `clusteradm` 
- Update default bundle to install OCM v0.16.1


pkg/cmd/init/cmd.go
```diff
@@ -74,9 +74,11 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 		"The credential file is the docker config json file and will be filled into the default image pull secret named open-cluster-management-image-pull-credentials.")
 	cmd.Flags().StringVar(&o.bundleVersion, "bundle-version", "default",
```
mikeshng, line 75: Is the `bundle-version` only useful for setting things to `latest`? For example, if I choose v0.16.1, it also sets other components (multicluster_controlplane, etc.) to v0.16.1, even though that version might not exist for those components.

mikeshng (commented):
Thank you for your PR!

I'm okay with the changes, but just wondering is the original `bundle-version` param only useful for setting `latest`? Should we consolidate the new param and the existing one? Or we should just add the extra parameter like you've done?

/assign @qiujian16

pkg/version/version.go
```diff
@@ -36,19 +38,58 @@ func Get() version.Info {
...
+	}
+	if bundle.MulticlusterControlplane == "" {
+		missingKeys = append(missingKeys, "multicluster_controlplane")
+	}
+	if len(missingKeys) > 0 {
+		return VersionBundle{}, fmt.Errorf(
+			"invalid version bundle file: missing required keys: [%s]",
+			strings.Join(missingKeys, ", "),
+		)
+	}
```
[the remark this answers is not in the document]
TylerGillson: Totally open to making that change if that's the OCM team's preference!

pkg/version/version.go
```diff
@@ -36,19 +38,58 @@ func Get() version.Info {
...
+	if bundle.AppAddon == "" {
+		missingKeys = append(missingKeys, "app_addon")
+	}
+	if bundle.PolicyAddon == "" {
+		missingKeys = append(missingKeys, "policy_addon")
+	}
+	if bundle.MulticlusterControlplane == "" {
+		missingKeys = append(missingKeys, "multicluster_controlplane")
+	}
+	if len(missingKeys) > 0 {
```
qiujian16, line 78: I think if a key is missing, we should fill it with the default value.

PR #509, opened 2025-08-18
✨ allow labeling resources created during addon create and enable

arturshadnik:

## Summary

This PR adds a new `--labels` flag to the `clusteradm addon create` and `clusteradm addon enable` commands. The flag takes comma-separated list of key=value pairs that will be added as labels to the AddOn resources created by the commands. 

Even though `enable` already supports annotations, labels are more useful when filtering resources, so I propose supporting both for that command.

## Summary by CodeRabbit

- **New Features**
  - Added a --labels flag to addon create and enable commands to attach multiple key=value labels to created resources; invalid label formats produce a clear error.
- **Bug Fixes**
  - Ensures label changes are persisted when updating an existing add-on.
- **Tests**
  - Added end-to-end checks verifying labels propagate to management, template, and managed-cluster resources.


TylerGillson (commented):
Please update the addon E2E test to include usage of this new flag. The filename `addon_create_test.go` is also not quite accurate. Maybe rename to `addon_lifecycle_test.go`?

PR #548, opened 2026-04-27
:bug: best-effort create cluster-info when missing on managed cluster join

rokej:

## Summary

When registering an OpenShift cluster as a managed cluster, clusteradm join can print this warning:

```
W0131 11:43:03.072250   76891 exec.go:204] Failed looking for cluster endpoint for the registering klusterlet: configmaps "cluster-info" not found
```

This is just a warning message and the join process continues successfully but as https://github.com/open-cluster-management-io/ocm/issues/1499 indicates, it is confusing. 

- When discovering the managed-cluster API URL for the klusterlet fails, the message no longer says “Failed looking for cluster endpoint…”.

- If the failure is ConfigMap cluster-info not found, the warning explains that kube-public/cluster-info is missing, and that join still continues without an external API URL for the klusterlet.

- For any other error, it logs the same leading sentence and appends : <error>.

- Those two cases are implemented as one klog.Warningf using a detail string (fmt.Sprintf(": %v", err) vs. the longer explanation for NotFound).

## Related issue(s)

Fixes # https://github.com/open-cluster-management-io/ocm/issues/1499

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved join robustness: when an external API server URL can't be retrieved during cluster join, the process now logs a clearer warning and continues, reducing interrupted setups and making join operations more resilient.

[Review Change Stack](https://app.coderabbit.ai/change-stack/open-cluster-management-io/clusteradm/pull/548)

### DIFF
--- a/pkg/cmd/join/exec.go
+++ b/pkg/cmd/join/exec.go
@@ -246,7 +246,8 @@ func (o *Options) complete(cmd *cobra.Command, args []string) (err error) {
 
 	klusterletApiserver, err := sdkhelpers.GetAPIServer(kubeClient)
 	if err != nil {
-		klog.Warningf("Failed looking for cluster endpoint for the registering klusterlet: %v", err)
+		detail := fmt.Sprintf(": %v", err)
+		klog.Warningf("Join continues without an external API server URL for the klusterlet because %s", detail)
 		klusterletApiserver = ""
 	} else if !preflight.ValidAPIHost(klusterletApiserver) {
 		klog.Warningf("ConfigMap/cluster-info.data.kubeconfig.clusters[0].cluster.server field [%s] in namespace kube-public should start with http:// or https://", klusterletApiserver)


pkg/cmd/join/exec.go
```diff
@@ -688,6 +695,77 @@ func (o *Options) createExternalBootstrapConfig() clientcmdapiv1.Config {
...
+	}
+}
+
+func loadCurrentCluster(config *clientcmdapi.Config) (*clientcmdapi.Cluster, error) {
+	currentCtx, exists := config.Contexts[config.CurrentContext]
+	if !exists {
+		return nil, fmt.Errorf("failed to find current context in kubeconfig")
+	}
+
+	currentCluster, exists := config.Clusters[currentCtx.Cluster]
```
zhujian7: The cluster-info configmap is created from the kubeconfig used for clusteradm join. If the user connects via a proxy, tunnel, or port-forward, the stored URL won't reflect the cluster's actual API server, right? — and this flows into ManagedCluster.spec.managedClusterClientConfigs.url on the hub.

rokej: @zhujian7 Thank you for reviewing it. You are absolutely right and I think there is no good source of truth for the API server URL.

rokej: @zhujian7 Although it is not necessarily always the correct server API URL from the hub cluster, it "predictably" publishes what the API URL used to join the hub. In my opinion, using this API URL is still good default for many clusters. ManagedCluster.spec.managedClusterClientConfigs.url on the hub should be informational.

zhujian7: Thanks @rokej for the fix on the warning handling. One concern — kube-public/cluster-info is a cluster infrastructure resource normally created by kubeadm/the installer, and is readable by all users including unauthenticated ones. I'm not sure clusteradm join should be auto-creating it — if the user connects via proxy/tunnel/port-forward, we'd publish a wrong URL into a world-readable immutable ConfigMap. In some cases empty is better than wrong.
cc @qiujian16 WDYT?

qiujian16: yeah I agree, I do not think we need to auto it if it is not correct.

rokej: @zhujian7 @qiujian16 Thank you for reviewing it. I agree. I will just update the warning message to indicate there is no failure.

rokej:
@coderabbitai review

PR #553, opened 2026-05-19
:bug: Use matching bundle version in e2e tests instead of latest

### LINKED ISSUE #552: e2e tests with BUNDLE_VERSION=latest use mismatched chart and images causing failures on release update PRs
**Describe the bug**

The clusteradm e2e test infrastructure uses `BUNDLE_VERSION=latest` for PRs targeting `main`. This pulls container images tagged `:latest` from quay.io (built from OCM `main` branch) but deploys the Helm chart vendored from the Go module dependency in `go.mod` (which can be a release tag like `v1.3.1` from a release branch). This creates an inherent chart-image version mismatch that breaks e2e tests whenever OCM `main` diverges from the vendored module version.

This makes it impossible to land PRs that update the OCM dependency to a release tag (e.g. v1.3.0 → v1.3.1) if OCM `main` has moved ahead with changes that require corresponding chart updates (new RBAC entries, new containers, new CRDs, etc.).

For example, https://github.com/open-cluster-management-io/ocm/pull/1494 added a new secret informer on `main` along with its required `resourceNames` RBAC entry. When clusteradm vendors OCM v1.3.1 (which does not include that PR), the e2e test deploys the old chart but runs the new image, causing RBAC failures and a timeout.

**To Reproduce**

1. Open a PR on clusteradm that updates `go.mod` to a release tag (e.g. `open-cluster-management.io/ocm v1.3.1`)
2. CI runs e2e tests with `BUNDLE_VERSION=latest`
3. The `:latest` images (from OCM `main`) expect RBAC/config that only exists in the `main` branch chart
4. The vendored chart (from the release tag) does not have those changes
5. e2e tests fail due to the mismatch

**Expected behavior**

The e2e test infrastructure should use a consistent pair of chart + images. Either:
- `BUNDLE_VERSION=latest` should use BOTH the chart and images from OCM `main`, or
- The e2e tests for release-update PRs should use a bundle version matching the vendored release (e.g. `BUNDLE_VERSION=1.3.1`), or
- The CI should detect that the vendored chart is behind `:latest` images and skip/adjust accordingly

**Environment**

**Additional context**

mikeshng:
## Summary

Use matching bundle version in e2e tests instead of latest

The e2e CI previously set BUNDLE_VERSION=latest for PRs targeting main, which pulls :latest container images (built from OCM main) but deploys the Helm chart vendored from the Go module dependency (a release tag). This chart-image version mismatch causes spurious RBAC and config failures whenever OCM main introduces chart changes not present in the vendored release, blocking unrelated PRs like version bumps.

The early-detection signal this was meant to provide was inaccurate: it tested a combination (release chart + main images) that no real deployment would ever use, so its failures did not represent real bugs.

Switch to BUNDLE_VERSION=default for all branches so the e2e always tests with a consistent set of chart, images, and compiled code matching the vendored OCM version.

## Related issue(s)

Fixes https://github.com/open-cluster-management-io/clusteradm/issues/552

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -64,6 +64,4 @@ jobs:
       with:
         go-version: ${{ env.GO_VERSION }}
     - name: E2E Tests
-      run: |
-        BUNDLE_VERSION=$([[ "${{ github.base_ref }}" == "main" ]] && echo latest || echo default) \
-        make test-e2e
+      run: make test-e2e
--- a/test/e2e/e2e-test.mk
+++ b/test/e2e/e2e-test.mk
@@ -7,7 +7,7 @@ export MANAGED_CLUSTER1_NAME := ${PROJECT_NAME}-e2e-test-c1
 export HUB_CTX := kind-${HUB_NAME}
 export MANAGED_CLUSTER1_CTX := kind-${MANAGED_CLUSTER1_NAME}
 
-BUNDLE_VERSION?=latest
+BUNDLE_VERSION := default
 
 clean-e2e: 
 	kind delete cluster --name ${HUB_NAME}


test/e2e/e2e-test.mk
```diff
@@ -8,6 +8,24 @@ export HUB_CTX := kind-${HUB_NAME}
...
+	@tmpdir=$$(mktemp -d) && \
+	git clone --depth=1 --branch main --filter=blob:none --sparse $(OCM_REPO) "$$tmpdir" && \
+	git -C "$$tmpdir" sparse-checkout set deploy/cluster-manager/chart deploy/klusterlet/chart && \
+	rm -rf $(OCM_CHART_VENDOR)/cluster-manager/chart/cluster-manager && \
+	cp -r "$$tmpdir/deploy/cluster-manager/chart/cluster-manager" $(OCM_CHART_VENDOR)/cluster-manager/chart/cluster-manager && \
+	rm -rf $(OCM_CHART_VENDOR)/klusterlet/chart/klusterlet && \
+	cp -r "$$tmpdir/deploy/klusterlet/chart/klusterlet" $(OCM_CHART_VENDOR)/klusterlet/chart/klusterlet && \
+	rm -rf "$$tmpdir"
+endif
+.PHONY: sync-latest-charts
```
dhaiducek: Seems like using Go commands would be simpler and less error-prone?

But I also feel like the core of the issue is that the `latest` tag refers to dev images whereas it usually refers to the latest version. This is going to go out and if users happen to use `latest`, they're still going to see the issue, so it seems we're doing hacky things just to appease the CI here?

```suggestion
sync-latest-charts:
ifeq ($(BUNDLE_VERSION),latest)
	go get open-cluster-management.io/ocm@main
	go mod vendor
endif
.PHONY: sync-latest-charts
```

mikeshng: This is what I briefly mentioned in Slack. I don't want to compile against main. I want it to still compile against whatever OCM version it is current set to in the go.mod. If I do what you suggested, the PR is building one thing but e2e is building another.

I do agree it seems hacky. Let me change the approach entirely.

.github/workflows/tests.yaml
```diff
@@ -65,5 +65,4 @@ jobs:
         go-version: ${{ env.GO_VERSION }}
     - name: E2E Tests
       run: |
-        BUNDLE_VERSION=$([[ "${{ github.base_ref }}" == "main" ]] && echo latest || echo default) \
-        make test-e2e
+        BUNDLE_VERSION=default make test-e2e
```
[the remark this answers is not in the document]
mikeshng: I've updated the PR title and summary already. You are looking at the old stuffs.

test/e2e/e2e-test.mk
mikeshng: Updated the PR title and summary to reflect the new approach. 

It's the only way to be consistent IMO for e2e.

TylerGillson (approved):

test/e2e/e2e-test.mk
```diff
@@ -7,7 +7,7 @@ export MANAGED_CLUSTER1_NAME := ${PROJECT_NAME}-e2e-test-c1
 export HUB_CTX := kind-${HUB_NAME}
 export MANAGED_CLUSTER1_CTX := kind-${MANAGED_CLUSTER1_NAME}
 
-BUNDLE_VERSION?=latest
+BUNDLE_VERSION := default
```
qiujian16, line 10: there will be an issue with this, since e2e will not use the latest to run test.

mikeshng: Yes, I should undo this.

mikeshng: https://github.com/open-cluster-management-io/clusteradm/pull/555

PR #562, opened 2026-07-01
:bug: fix: propagate `clusteradm` flags into Helm at initialization time

### LINKED ISSUE #561: Chart installs with `install hub-addon` don't respect `clusteradm` flags
**Describe the bug**

Running `clusteradm install hub-addon --dry-run` logs it's in dry run mode and then applies manifests anyway.

**To Reproduce**

```
clusteradm install hub-addon --names argocd --dry-run
```

**Expected behavior**

Nothing deployed during dry run

**Environment ie: OCM version, clusteradm version, Kubernetes version and provider:**

`clusteradm v1.3.1`
Kubernetes v1.35 with `kubectl` client v1.36 using a Kind cluster

**Additional context**

dhaiducek:
Previously this was setting the `dryRun` in the `values.yaml`, which was not propagating to the Helm logic. This also adds logic to not print manifests outside of the dry run.

This also places manifest printing behind the dry run conditional to reduce noise.

Closes #561

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Helm dry-run handling so it’s applied consistently across install flows.
  * Reduced dry-run output noise by showing Helm release manifests only during dry-run.
  * Applied namespace-creation behavior once per installation to avoid inconsistent results.
* **Tests**
  * Added an integration test verifying dry-run hub add-on install does not create the target namespace.
  * Expanded integration CRD test paths to cover additional CRD versions.

### DIFF
--- a/pkg/helpers/helm/helm.go
+++ b/pkg/helpers/helm/helm.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"gopkg.in/yaml.v2"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 
 	"github.com/gofrs/flock"
 	"github.com/pkg/errors"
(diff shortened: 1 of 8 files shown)


pkg/helpers/helm/helm.go
```diff
@@ -45,6 +48,17 @@ func NewHelm() *Helm {
...
+	h.restClientGetter = &restClientGetter
+}
+
+func (h *Helm) restClientGetterOrDefault(restClientGetter *genericclioptions.RESTClientGetter) genericclioptions.RESTClientGetter {
+	if restClientGetter != nil {
+		return *h.restClientGetter
+	}
+	return h.settings.RESTClientGetter()
+}
+
```
dhaiducek: These new helpers are necessary for testing since Helm by default uses the default kubeconfig, and tests need to be able to point at the envtest environment instead. One alternative might be to write a kubeconfig file and point to it, but I think the approach here might be cleaner/less hacky.

dhaiducek: Nevermind. I realized this logic flow was necessary on a wider basis.

PR #563, opened 2026-07-02
:sparkles: feat: add `raw` option for token fetch

### LINKED ISSUE #315: get token: Print only the token in ouptut
# Feature Request

Allow `clusteradm get token` to only print the token, and/or write just the token to file.
This will help developers/admin script the remote cluster registration as follows (for KinD):

```sh
kubectl config use-context ${CTX_HUB_CULSTER}
clusteradm init --wait --context ${CTX_HUB_CLUSTER}
hubAPIServer=$(kubectl config view --minify --output jsonpath="{.clusters[*].cluster.server}")
hubToken=$(clusteradm get token --context ${CTX_HUB_CLUSTER})
clusteradm join \
    --hub-token ${hubToken} \
    --hub-apiserver ${hubAPIServer} \
    --wait \
    --cluster-name ${MANAGED_CLUSTER} \    
    --force-internal-endpoint-lookup \
    --context ${CTX_MANAGED_CLUSTER}
```

dhaiducek:
Adds an option to print only the token when running `get token`

Closes #315

## Summary by CodeRabbit

* **New Features**
  * Added `raw` as an accepted `--output` option for the token command.
  * Updated `--output` help text to list `json`, `text`, and `raw`.
* **Bug Fixes**
  * Token output now matches the selected format precisely (`json`, `text`, or `raw`).
  * Empty tokens now return an error instead of silently printing “token doesn’t exist”.
  * Invalid `--output` values now return an error instead of falling back to text.

### DIFF
--- a/pkg/cmd/get/token/cmd.go
+++ b/pkg/cmd/get/token/cmd.go
@@ -46,7 +46,7 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 
 	cmd.Flags().StringVar(&o.outputFile, "output-file", "", "The generated resources will be copied in the specified file")
 	cmd.Flags().BoolVar(&o.useBootstrapToken, "use-bootstrap-token", false, "If set then the bootstrap token will used instead of a service account token")
-	cmd.Flags().StringVarP(&o.output, "output", "o", "text", "output should be json or text")
+	cmd.Flags().StringVarP(&o.output, "output", "o", "text", "output should be json, text, or raw")
 
 	return cmd
 }
--- a/pkg/cmd/get/token/exec.go
+++ b/pkg/cmd/get/token/exec.go
@@ -66,7 +66,9 @@ func (o *Options) writeResult(token, host string) error {
 		fmt.Println(o.Streams.Out, "token doesn't exist")
 		return nil
 	}
-	if o.output == "json" {
+
+	switch o.output {
+	case "json":
 		err := clusteradmjson.WriteJsonOutput(o.Streams.Out, clusteradmjson.HubInfo{
 			HubToken:     token,
 			HubApiserver: host,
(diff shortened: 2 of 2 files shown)


mikeshng (approved):
/lgtm

/assign @TylerGillson

dhaiducek:
Thanks for the review @mikeshng! Pushed a no-op to re-run the flaky E2E:

```
  [FAILED] Timed out after 300.000s.
  Expected success, but got an error:
      <*errors.errorString | 0x232123e82ca0>: 
      not all clusters are deleted
      {
          s: "not all clusters are deleted",
      }
  In [BeforeEach] at: /home/runner/work/clusteradm/clusteradm/test/e2e/util/helper.go:81 @ 07/03/26 01:24:05.405
```

dhaiducek:
This note from CodeRabbit is also interesting. I'm not sure it should be changed here, but something to note. I'm not even sure the token would ever be empty here since it seems like an error would be returned prior to that, but maybe this should also return an error instead of writing to stdout?

pkg/cmd/get/token/exec.go (1)> `65-68`: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_
> **`fmt.Println(o.Streams.Out, ...)` doesn't write to `o.Streams.Out`.**
> `fmt.Println` treats all arguments as values to print, not a destination writer — this prints the Writer's value representation to the real stdout instead of writing "token doesn't exist" to `o.Streams.Out`. Since this path guards all three new output modes (json/text/raw), the empty-token case is broken for every mode.
> 
> 🐛 Proposed fix
> ```diff
>  	if len(token) == 0 {
> -		fmt.Println(o.Streams.Out, "token doesn't exist")
> +		fmt.Fprintln(o.Streams.Out, "token doesn't exist")
>  		return nil
>  	}
> ```

dhaiducek:
Updated:
- Removed trailing newline from returned raw output
- Updated to make an empty token an error rather than printing the "token is empty" string

PR #565, opened 2026-07-06
:bug: fix: update broken URLs

dhaiducek:
Also prefer `Fprintln` when there is no formatted string.

## Summary by CodeRabbit

* **Documentation**
  * Updated several command help and status messages to point to the latest getting-started and scenario guides.
  * Improved link accuracy for cluster-proxy and cluster-join next-step guidance.

* **Bug Fixes**
  * Standardized terminal output across cluster management commands so status and warning messages display cleanly with consistent line breaks.


mikeshng (approved):
/lgtm

Thank you for this fix!

PR #566, opened 2026-07-07
:bug: fix: use pointer in Helm client for `clusteradm` flags

dhaiducek:
The previous test passed because I was passing the dry run boolean in directly. In reality, the options are initialized at build time, so the Helm client needs to be a pointer so that the dry run updates at run time.

Followup to:
- #562

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Helm dry-run handling so install operations now consistently respect the selected dry-run setting.
  * When dry-run is enabled, Helm now uses client-only behavior and prints the expected release manifest output.


mikeshng (approved):
/lgtm

Thanks for the fix Dale!

PR #575, opened 2026-09-16
:bug: Reject --clusters in proxy kubectl instead of silently ignoring it

### LINKED ISSUE #366: clusteradm proxy kubectl does not allow multiple clusters
`clusteradm proxy kubectl` takes in a generic `ClusterOption` which includes both the option to specify a single cluster or multiple clusters (`--clusters`), but the code seems to only accept a single cluster https://github.com/open-cluster-management-io/clusteradm/blob/329f6fef88bfac77ac2291e1696f5615b12288e3/pkg/cmd/proxy/kubectl/cmd.go#L74. This should be explicitly stated as a limitation and not appear in the `--help` output of the command, should be supported, or there should be a clearer error message ("resource name may not be empty").

magic-peach:
## Summary

proxy kubectl only ever used o.ClusterOption.Cluster, so passing --clusters instead of --cluster silently proceeded with an empty cluster name and failed later with a confusing apiserver error. validate() now rejects --clusters explicitly and requires --cluster, and --clusters is hidden from this command's help since it never worked here. The shared ClusterOption type and other commands that do support multiple clusters are unaffected.

## Related issue(s)

Fixes #366

## Summary by CodeRabbit

* **Documentation**
  * Clarified single-cluster usage through the `--cluster` option.
  * Documented the managed service account token certificate requirement.
  * Removed the multi-cluster option from help output.

* **Bug Fixes**
  * Improved validation for missing or conflicting cluster settings.
  * Added clearer errors when managed service account credentials are missing.


pkg/cmd/proxy/kubectl/cmd.go
```diff
@@ -173,6 +173,7 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	}
 
 	o.ClusterOption.AddFlags(cmd.Flags())
+	_ = cmd.Flags().MarkHidden("clusters")
```
dhaiducek, line 176: Oh cool--I didn't know about this one!

pkg/cmd/proxy/kubectl/cmd.go
```diff
@@ -43,7 +43,7 @@ func NewCmd(clusteradmFlags *genericclioptionsclusteradm.ClusteradmFlags, stream
 	cmd := &cobra.Command{
 		Use:   "kubectl",
 		Short: "Use kubectl through cluster-proxy addon.",
-		Long:  "Use kubectl through cluster-proxy addon. (Only supports managed service account token as certificate.)",
+		Long:  "Use kubectl through cluster-proxy addon. (Only supports managed service account token as certificate.) Only a single cluster is supported per invocation, use --cluster.",
```
dhaiducek: Wording nit

```suggestion
		Long:  "Use kubectl with the specified managed cluster through cluster-proxy addon. (Supports only managed service account token as certificate, and only a single cluster per invocation specified using --cluster.)",
```

dhaiducek (commented):
LGTM! Thanks for the bug fix to aid in the usability of `clusteradm`!

I have a minor wording nit for the description to consider to put all of the command requirements inside of the parentheses.

magic-peach:
@dhaiducek could you please check?

dhaiducek (approved):
LGTM! Thanks for the usability fix!

PR #578, opened 2026-09-18
:seedling: Add unit tests for the check package

magic-peach:
## Summary

pkg/helpers/check/check.go has three preflight functions, CheckForHub, CheckForManagedCluster, and CheckForKlusterletCRD, used by clusteradm_flags.go and unjoin/exec.go, plus IsFeatureEnabled used by get/hubinfo/exec.go. None of these had any test coverage.

This adds a table driven test file with subtests covering the found, not found, and error cases for each function.

## Related issue(s)

None

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of environment validation and unjoin checks when confirming required API resources.

* **Tests**
  * Added unit test coverage for resource availability checks across present, missing, and unavailable API resources.
  * Added coverage for feature-gate behavior, including enabled, disabled, empty, and unrecognized features.


dhaiducek (changes requested):
> 🧹 Nitpick comments (1)
> > pkg/helpers/check/check_test.go (1)> `56-58`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Cover non-NotFound discovery errors.** `FakeDiscovery.ServerResourcesForGroupVersion` returns a NotFound error when `Resources` is `nil`, so these fixtures cover only the NotFound cases. Because the tests assert only `wantErr`, they would also pass if a helper returned the wrong non-NotFound error behavior.
> > > Add a non-NotFound discovery-error case to each table and assert the helper-specific contract:
> > > 
> > > * `CheckForHub` formats the discovery error with the group version.
> > > * `CheckForManagedCluster` returns its formatted error without the original error.
> > > * `CheckForKlusterletCRD` returns the original discovery error.
> > > 

Thanks for these tests! More unit test coverage is certainly welcome! This CodeRabbit nit is relevant since it'd serve as a more fine-tuned test.

Also this is a small change that adds a lot to our `vendor/` directory. It would be simplified if the parent function signatures accepted `discovery.DiscoveryInterface` since all of them call `discovery.Discovery()`. Then the new fake client should no longer be necessary.

magic-peach:
Thanks for the feedback @dhaiducek . Changed CheckForHub, CheckForManagedCluster, and CheckForKlusterletCRD to take discovery.DiscoveryInterface instead of the full clientset, so tests use a plain fake discovery client and no longer need the vendored fake clientsets. Also added the non-NotFound discovery error case to each table, asserting the specific contract per function (group version included, original error dropped, original error preserved).

PR #579, opened 2026-09-18
:bug: fix WatchUntil's misleading error message on a closed watch channel

magic-peach:
## Summary

WatchUntil already correctly ignores non-matching events and keeps waiting for a matching one, at all 5 call sites (including join and delete). The only actual failure path is the watch channel closing before the condition is met, but the error it returned in that case was "unexpected watch event received", which is misleading since no unexpected event was ever received. This sent at least one issue reporter chasing the wrong cause.

Fixed the error message to describe the real failure (the watch channel closed before the condition was satisfied), and added the first unit test coverage for this shared helper: matching condition succeeds, non-matching events are ignored until the channel closes, and a watchFunc error is returned without starting a watch.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the error message shown when monitoring ends before the expected event is received.

- **Tests**
  - Added coverage for successful event matching, early watch closure, and watch-creation errors.


dhaiducek:
Thanks--seems like a fair change. However, I've detached this PR from the issue since the core of the issue is not resolved in this PR. I've been working on a fix there, so I've assigned the issue to myself.

